PR #38, opened 2021-04-15
Willem/committee add constituents

### LINKED ISSUE #36: Implement constituent committee to `Committee` Pallet
## Implementation details

PINT governance will be administered in two layers: the PINT Council and the **Constituent Committee**. The PINT Council will consist of 7 representatives, while the **Constituent Committee** will consist of one representative from each project that has assets included in the index.

The Constituent Committee retains veto power over the decisions of the PINT Council. If the majority of the Constituent Committee votes "nay" following a PINT Council Governance vote, the action will fail. If there is no majority "nay" vote from the Constituent Committee, the action will pass.

## Acceptance Criteria

- Second layer of governance to be included
- veto power
- council selection

willemolding:
## Changes

Adds the two-layer governance (council and constituents) to the committee pallet.

- Members now have an associated membership type and are stored in a StorageMap
- initialize_members adds initial members as council
- change_members allows for adding constituent members only
- Refactor of VotesAggregate to store a single vec of votes rather than three for ayes, nays and abstains
- Change vote acceptance criteria to allow the constituent members to veto
- Adds type parameter for MinCouncilVotes required to pass an action
- Adds additonal error types for the different reasons a proposal can be rejected
- Refactor tests

## Tests

```
cargo test
```

New tests added for constituent veto logic

## Issues
- Closes #36

### DIFF
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -403,9 +403,9 @@ pub mod pallet {
     /// The existing council can then vote to add them as concil members
     impl<T: Config> ChangeMembers<AccountIdFor<T>> for Pallet<T> {
         fn change_members_sorted(
-            _incoming: &[AccountIdFor<T>],
+            incoming: &[AccountIdFor<T>],
             outgoing: &[AccountIdFor<T>],
-            new: &[AccountIdFor<T>],
+            _sorted_new: &[AccountIdFor<T>],
         ) {
             // Remove outgoing members from any currently active votes
             for proposal_hash in ActiveProposals::<T>::get() {
@@ -415,8 +415,10 @@ pub mod pallet {
                     }
                 });
             }
-            Members::<T>::drain().for_each(drop); // remove all the members from the map
-            for member in new {
+            for leaving_member in outgoing {
+                Members::<T>::remove(leaving_member)
+            }
+            for member in incoming {
                 Members::<T>::insert(member, MemberType::Constituent);
             }
         }
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -98,8 +98,9 @@ pub mod pallet {
         StorageMap<_, Blake2_128Concat, HashFor<T>, (), OptionQuery>;
 
     #[pallet::storage]
-    /// Stores a vector of account IDs of current committee members
-    pub type Members<T: Config> = StorageValue<_, Vec<AccountIdFor<T>>, ValueQuery>;
+    /// Maps accountIDs to their member type (council or constituent)
+    pub type Members<T: Config> =
+        StorageMap<_, Blake2_128Concat, AccountIdFor<T>, MemberType, OptionQuery>;
 
     #[pallet::storage]
     /// Store a mapping (hash) -> VoteAggregate for all existing proposals.
@@ -121,7 +122,7 @@ pub mod pallet {
         Proposed(AccountIdFor<T>, T::ProposalNonce, T::Hash),
         /// A vote was cast
         /// \[voter_address, proposal_hash, vote\]
-        VoteCast(AccountIdFor<T>, T::Hash, Vote),
+        VoteCast(CommitteeMember<AccountIdFor<T>>, T::Hash, Vote),
         /// A proposal was closed and executed. Any errors for calling the proposal action
         /// are included
         /// \[proposal_hash, result\]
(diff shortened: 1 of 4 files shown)


pallets/committee/src/types.rs
```diff
@@ -66,33 +109,61 @@ impl<AccountId: Default + PartialEq, BlockNumber: Default> VoteAggregate<Account
...
+            .filter(|x| &x.member.account_id == voter)
+            .count()
+            > 0
     }
 
-    pub fn participants(&self) -> usize {
-        self.ayes.len() + self.nays.len() + self.abstentions.len()
+    /// produce a tuple of the vote totals: (ayes, nays, abstentions)
+    /// Can optionally filter by membership type to only tally council or constituent votes
+    pub fn tally(&self, member_type: &Option<MemberType>) -> (usize, usize, usize) {
```
mattsse, line 131: With optional input arguments, I'm usually not sure which variant is more appropriate:

`&Option<T>` or `Option<&T>`, while you can turn `Option<T>` easily into the later with `as_ref`, turning a `&T` into `&Option<T>` would require taking ownership if I'm not mistaken.

However I don't think that this decision has any impact here.

pallets/committee/src/types.rs
```diff
@@ -33,29 +74,31 @@ pub enum CommitteeOrigin<AccountId, BlockNumber> {
...
         end: BlockNumber,
     ) -> Self {
-        Self {
-            ayes,
-            nays,
-            abstentions,
-            end,
-        }
+        let votes = std::iter::empty()
+            .chain(ayes.into_iter().map(|x| x.into_vote(Vote::Aye)))
```
mattsse, line 97: I think the `std::iter::empty` is not required here

```suggestion
        let votes = ayes.into_iter().map(|x| x.into_vote(Vote::Aye))
```

pallets/committee/src/types.rs
```diff
@@ -22,6 +22,47 @@ impl<T: Config> Proposal<T> {
...
+            account_id,
+            member_type,
+        }
+    }
+
+    pub fn into_vote(self, vote: Vote) -> MemberVote<AccountId> {
+        MemberVote { member: self, vote }
+    }
+}
+
```
mattsse, line 53:
```suggestion
/// A committee member together with their cast vote.
```

pallets/committee/src/lib.rs
```diff
@@ -359,23 +382,30 @@ pub mod pallet {
...
     }
 
+    /// Initialize council members. Can only be done once.
+    /// Constituent members must be initialized later by voting by the council
     impl<T: Config> InitializeMembers<AccountIdFor<T>> for Pallet<T> {
         fn initialize_members(members: &[AccountIdFor<T>]) {
             if !members.is_empty() {
                 assert!(
-                    Members::<T>::get().is_empty(),
+                    Members::<T>::iter().count() == 0,
```
mattsse, line 391: Just spotted this simplification
```suggestion
                    assert_eq!(Members::<T>::iter().count(), 0, "Members are already initialized!");
```

mattsse (changes requested):
I only noticed a few minor things

pallets/committee/src/types.rs
```diff
@@ -66,33 +109,61 @@ impl<AccountId: Default + PartialEq, BlockNumber: Default> VoteAggregate<Account
...
+        self.votes
+            .retain(|x| !voters.contains(&x.member.account_id));
     }
 
     pub fn has_voted(&self, voter: &AccountId) -> bool {
-        self.ayes.contains(voter) | self.nays.contains(voter) | self.abstentions.contains(voter)
+        self.votes
+            .iter()
+            .filter(|x| &x.member.account_id == voter)
+            .count()
```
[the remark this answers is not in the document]
willemolding: Sweet :+1:

pallets/committee/src/types.rs
willemolding: definitely not required but I prefer it for readability. I'm fairly sure it results in the same thing when compiled.

pallets/committee/src/types.rs
willemolding: Yeah I'm also not sure in this case. I switched to `Option<&T>`. I think passing an option that holds a reference to T makes more sense than passing a reference to an option that owns T

pallets/committee/src/types.rs
mattsse: You're right it is actually easier to read that way, I like that the relevant iterators are all aligned the same way.

mattsse (approved):

PR #39, opened 2021-04-15
fix(deps): upstream rococo-v1 

mattsse:
## Changes

- replace `ModuleId` with `PalletId` to address https://github.com/paritytech/substrate/pull/8372 
- replace every import of soon to be deprecated `Module` in `construct_runtime!` with `Pallet`
- update every `system::Config` trait implementation
- update node and runtime to latest version of the parachain-template https://github.com/substrate-developer-hub/substrate-parachain-template/commit/40b858149b212e493da08d80e0fc5b06a6b0b72d
- update ci: use newer ubuntu and nightly toolchain

~~The runtime remains unchanged and currently doesn't compile, the templates aren't updated yet, I wasn't able to fix this. XCM will also probably change quite a bit.~~

## Tests

```
cargo test
```

## Issues

- closes #37

PR #40, opened 2021-04-15
Add price feed pallet

### LINKED ISSUE #30: Integrate oracle pallet and add price feed
In order to exchange assets for PINT, the value of those assets as well as PINT’s NAV needs to be computed. To achieve this a price-feed service is required.

mattsse:
## Changes

Blocked by #39 because the substate-v3 chainlink feed depends on a newer substrate rococo revision
- Adds skeleton of price-feed pallet
- Integrates the chainlink-feed pallet via `FeedOracle`
- Adds `AssetPricePair` type to represent a base/quote price pair
- Adds `PirceFeed` trait to fetch asset pairs

~~The chainlink pallet is abstracted via the `FeedOracle` trait, the PINT price-feed pallet will be an additional abstraction on top of it. While the design for the price-feed isn't final, I believe we probably will need some storage items, therefor the price-feed will be a pallet instead of an additional trait that is implemented for `FeedOracle`~~

The `PriceFeed` trait is the intended abstraction on top the price feed pallet
New prices are inserting into the designated feeds via the existing chainlink feed pallet; At this point it is assumed that all feeds, including the PINT price feed will use a common base currency (DOT).
~~The price feed pallet is configured with a `Percision` which is applied to all price feeds, since feeds can have different `decimals`, a conversion will be required.~~

Update: 
* FixedU128 was chosen as type to represent price ratios.
* The admin needs to create a mapping between assetId and feedId.

## Tests

```
cargo test
```

## Issues

- closes #30


pallets/price-feed/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+        // TODO this needs to be factored in when converting the feed prices with their decimals
+        type Precision: PerThing + Encode;
+
+        /// Type used to identify the asset's feed.
+        type AssetFeedId: Parameter + Member;
+
+        /// The internal oracle that gives access to the asset's price feeds.
+        ///
+        /// NOTE: this assumes all the feed's provide data in the same base currency.
+        /// When querying the price of an asset (`quote`/`asset`) from the, its price is given by means of the asset pair `(base / quote)`. (e.g. DOT/PINT)
```
willemolding, line 50:
```suggestion
        /// When querying the price of an asset (`quote`/`asset`) from the oracle, its price is given by means of the asset pair `(base / quote)`. (e.g. DOT/PINT)
```

pallets/price-feed/src/types.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// Calculates the total volume of the provided units of the `quote` assetId w.r.t. price pair
+    pub fn volume<N>(&self, units: N) -> Price
+    where
+        N: Into<Price>,
+    {
+        self.price * Price::saturated_from(units)
+    }
+
+    /// Calculates the total volume of the provided units of the `base` assetId w.r.t. price pair
+    pub fn reciprocal_volume<N>(&self, units: N) -> u128
```
willemolding, line 52: It would be great it we could remove the concrete type here (`u128`) and use a more generic type (e.g. `Volume`)

pallets/price-feed/src/types.rs
```diff
@@ -0,0 +1,73 @@
...
+        N: Into<u128>,
+    {
+        self.price.saturating_reciprocal_mul(units.into())
+    }
+}
+
+impl<AssetId, Price> AssetPricePair<AssetId, Price>
+where
+    AssetId: Member,
+    Price: PerThing + From<u128>,
```
willemolding, line 63: see above

pallets/price-feed/src/types.rs
```diff
@@ -0,0 +1,73 @@
...
+{
+    /// Turns this price pair of `base/quote` into a price pair of `quote/base`
+    pub fn invert(self) -> Self {
+        Self {
+            base: self.quote,
+            quote: self.base,
+            price: self.price.saturating_reciprocal_mul(1).into(),
+        }
+    }
+}
```
willemolding, line 73: It would be great if this file could include a testing module for the type methods with full coverage

willemolding (changes requested):
I'm having trouble building this locally and also I think there is a `cargo fmt` error. 

New additions look great. Some more test would be good. The `feed_creation_should_work` is just a smoke test, we should probably at least test calling the `get_price` implementation. It would also be great to have test for the `AssetPricePair` methods.

pallets/price-feed/src/types.rs
mattsse: after further consideration, I opted for `[FixedU128](https://substrate.dev/rustdocs/v3.0.0/sp_arithmetic/fixed_point/struct.FixedU128.html)` as the price type, which offers precision for up to 18 digits. I think this is sufficient represent the price in DOT.

willemolding: Looks great

willemolding (approved):
Looks great! Happy to merge once you have sorted out the dependency issues

Issue #25, opened 2021-03-30
Implement `deposit` API call


dutterbutter:
Implement `deposit` API call detailed in the specification under AssetIndex Pallet.
## Implementation details

- Dispatches transfer Call to relay chain to move DOTs into the PINT Index account
- Fetches PINT NAV and DOT price from price feed
- Mints PINT proportional to DOT deposit
- Updates index to reflect deposit

|  Param 	| Type    	| Desc                          	|
|:------:	|---------	|-------------------------------	|
| amount 	| Balance 	| Amount of DOT being deposited 	|

mattsse:
Currently, the deposit of funds is expected to happen via cross-chain messages. A user must first send a certain amount of an asset, that is supported by PINT, from another chain to the equivalent PINT address via XCM. This will also require that PINT will the same SS58 prefix as its relay chain. As soon as the XCM arrives on PINT, corresponding tokens can be minted on PINT  ([downward transfer](https://wiki.polkadot.network/docs/en/build-parachains-rococo#downward-transfers)).
The deposited funds could be directly converted into PINT tokens based on the current exchange rate. However, there are possible scenarios in which an exchange might not be successful, e.g. there is currently no price pair available because an oracle for the currency is possibly locked.

Also XCM will takes a bit of time so when the call arrives on PINT the price (PINT/Asset) may have already changed. 
Therefore, an approach that supports a variety of balances for assets for each account would make sense, so that a deposit via XCM always increases the corresponding account balance of the user first. Only in a second step can assets deposited on PINT be invested put in the index (in exchange for PINT tokens), or possibly even exchanged for other assets first.
the same would happen if funds were withdrawn from the index. Thus, liquidating PINT tokens would first increase the balance of the user's chosen assets before the user can transfer them via XCM.

This multi-asset approach would look similar to [MultiCurrency](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/traits/src/currency.rs#L16) identified by an asset identifier.

PR #48, opened 2021-04-26
feat: add asset depository pallet

mattsse:
## Changes

- Adds an additional palette which sole purpose is to track asset balances for a user
- It's abstracted via the `MultiAssetDepository` trait, which is intended to be included in the asset transaction routine for depositing and withdrawing funds on a cross-chain message issued by a user

## Tests

```
cargo test
```

## Issues

- Adds the necessary depository functionality for #25 in order to deposit cross chain assets https://github.com/ChainSafe/PINT/issues/25#issuecomment-824926629 to temporarily store them before converting them to PINT or withdrawing them via XCM upon a user request

### DIFF
--- a/pallets/asset-depository/src/lib.rs
+++ b/pallets/asset-depository/src/lib.rs
@@ -16,7 +16,7 @@ mod types;
 // this is requires as the #[pallet::event] proc macro generates code that violates this lint
 #[allow(clippy::unused_unit)]
 pub mod pallet {
-    use crate::traits::MultiAssetDepository;
+    pub use crate::traits::MultiAssetDepository;
     use crate::types::AccountBalance;
     use frame_support::sp_runtime::traits::{CheckedAdd, CheckedSub};
     use frame_support::{
--- a/pallets/asset-depository/src/lib.rs
+++ b/pallets/asset-depository/src/lib.rs
@@ -105,7 +105,6 @@ pub mod pallet {
     }
 
     impl<T: Config> MultiAssetDepository<T::AssetId, AccountIdFor<T>, T::Balance> for Pallet<T> {
-
         /// The total amount of the given asset currently held
         fn aggregated_balance(asset_id: &T::AssetId) -> T::Balance {
             TotalBalance::<T>::get(asset_id)
--- a/pallets/asset-depository/src/lib.rs
+++ b/pallets/asset-depository/src/lib.rs
@@ -1,0 +1,151 @@
+// Copyright 2021 ChainSafe Systems
+// SPDX-License-Identifier: LGPL-3.0-only
+
+//! # Asset Depository Pallet
+//!
+//! Provides support for storing the balances of multiple assets
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub use pallet::*;
+
+mod traits;
+mod types;
+
+#[frame_support::pallet]
+// this is requires as the #[pallet::event] proc macro generates code that violates this lint
+#[allow(clippy::unused_unit)]
+pub mod pallet {
+    use crate::types::AccountBalance;
+    use frame_support::{
+        pallet_prelude::*,
+        sp_runtime::{
+            traits::{AtLeast32BitUnsigned, Zero},
+        }
+    };
+    use frame_system::pallet_prelude::*;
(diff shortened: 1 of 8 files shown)


pallets/asset-depository/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                    .checked_add(&amount)
+                    .ok_or(Error::<T>::TotalBalanceOverflow)?;
+
+                // SAFETY: this can't overflow because the balance for an account is
+                //  at most equal to the total balance and adding to it already succeeded
+                Self::set_available_balance(
+                    &asset_id,
+                    who,
+                    Self::available_balance(&asset_id, who) + amount,
+                );
```
clearloop, line 144: Why not using https://substrate.dev/rustdocs/v3.0.0/sp_arithmetic/traits/trait.Saturating.html#tymethod.saturating_add ?

pallets/asset-depository/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+                balance.available = balance
+                    .available
+                    .checked_sub(&amount)
+                    .ok_or(Error::<T>::NotEnoughBalance)?;
+                Ok(())
+            })?;
+
+            // SAFETY: this can't underflow because the total balance for an asset is at least
+            //  equal to the total balance in the given account and subtracting already succeeded
+            TotalBalance::<T>::mutate(asset_id, |total| *total -= amount);
```
clearloop, line 170: https://substrate.dev/rustdocs/v3.0.0/sp_arithmetic/traits/trait.Saturating.html#tymethod.saturating_sub

pallets/asset-depository/src/types.rs
```diff
@@ -0,0 +1,28 @@
...
+    /// this, but it is the total pool what may in principle be transferred,
+    /// reserved.
+    ///
+    /// This is the only balance that matters in terms of most operations on
+    /// tokens.
+    pub available: Balance,
+    /// Balance which is currently locked and can't be accessed by the user.
+    ///
+    /// This is intended to reserve an amount of this asset for PINT related operations, so that it can be spend.
+    pub reserved: Balance,
```
clearloop, line 20: To be confirmed, we don't need methods reversing balances in this PR?

pallets/asset-depository/src/traits.rs
```diff
@@ -0,0 +1,22 @@
...
+
+use frame_support::sp_runtime::DispatchResult;
+
+/// An abstraction over the multiple balances for different assets
+pub trait MultiAssetDepository<AssetId, AccountId, Balance> {
+    /// The total amount of the given asset currently held
+    fn aggregated_balance(asset_id: &AssetId) -> Balance;
+
+    /// The total balance of an asset of a user
+    fn total_balance(asset_id: &AssetId, who: &AccountId) -> Balance;
```
clearloop, line 12: So, if we want to get `reversed_balance` in the future, we prefer to use `who.total_balance().saturating_sub(who.available_balance())`?

clearloop (approved):
Except for the `unused imported` warning in `mock.rs`, LGTM

pallets/asset-depository/src/lib.rs
mattsse: if we were to use saturating, we would be guaranteed no error, but the information would be lost as to how much we are over, which would cause us to leak balance. This should definitely end in an error if we do not have enough more capacity to hold higher values.

pallets/asset-depository/src/lib.rs
mattsse: I decided against saturating operations because values higher than the maximum are simply capped and that is not our intention.

pallets/asset-depository/src/types.rs
mattsse: This could possibly become useful soon, once we figure out how to integrate everything with XCM and if we need to lock funds. But you are right at the moment it is not used, it's just a precaution at this point.

pallets/asset-depository/src/traits.rs
mattsse: Eventually we can add a `LockableAssetDepository` that will make use of `reserve`. My guess at this moment is, that we could use something like this when need to temporarily lock something because we're waiting on the price oracle or if we need to disable deposit/withdraw for certain assets momentarily.

Issue #31, opened 2021-04-07
RemoteAssetManager pallet


dutterbutter:
## Implementation details

Provides capabilities to bond/unbond and transfer assets on other chains. Utilized by the AssetIndex to interact with assets on other parachains.

## Acceptance Criteria

The completed pallet should have the following functionality:

- [x] `transfer`
- [x] `bond`
- [x] `unbond`

PR #50, opened 2021-04-27
feat(asset-index): implement deposit extrinsic

### LINKED ISSUE #25: Implement `deposit` API call

mattsse:
## Changes

- Implements the `AssetIndex::deposit` extrinsic which supports adding a given amount of a the given asset to the index in exchange of newly minted PINT tokens.
- The `deposit` function takes as input an `assetId` that identifies the asset for which the user wants to convert the given `amount` into PINT tokens.
- This will fail:
    - if the corresponding holding doesn't exist yet or the asset is not liquid.
    - No price pair (PINT/Asset) is available
    - The user's sovereign account can't cover the requested amount. 

## Tests

```
cargo test
```

## Issues

- Closes #25
- Related to #31


pallets/asset-index/src/lib.rs
```diff
@@ -121,6 +149,63 @@ pub mod pallet {
...
+            Ok(().into())
+        }
+
+        #[pallet::weight(10_000)] // TODO: Set weights
+        pub fn withdraw(
+            origin: OriginFor<T>,
+            _amount: BalanceFor<T>,
+        ) -> DispatchResultWithPostInfo {
+            let _caller = ensure_signed(origin)?;
+
```
willemolding, line 199:
```suggestion
todo!();
```

pallets/asset-index/src/lib.rs
```diff
@@ -121,6 +149,63 @@ pub mod pallet {
...
+            asset_id: T::AssetId,
+            amount: BalanceFor<T>,
+        ) -> DispatchResultWithPostInfo {
+            let caller = ensure_signed(origin)?;
+
+            let mut holding = Holdings::<T>::get(&asset_id)
+                .filter(|holding| matches!(holding.availability, AssetAvailability::Liquid(_)))
+                .ok_or(Error::<T>::UnsupportedAsset)?;
+
+            let price = T::PriceFeed::get_price(asset_id.clone())?;
```
willemolding, line 168: Consider adding a test for the failure cases:
- When a price feed isn't available
- When the various units overflow
- Interacting with the MultiAssetDepository fails

willemolding (approved):
It would be solid to have full test coverage of the extrinsic failure branches. Also a few clippy things here and there. Otherwise LGTM :+1:

PR #55, opened 2021-04-28
feat(asset-index): introduce MultiAssetAdapter

### LINKED ISSUE #51: Implement the `AssetTransactor` type required for the XcmConfig
## Issue summary

- The `AssetTransactor` will handle withdrawing and depositing assets upon received XCM
- This will require several conversions
    - from `MultiLocation` to `Asset`
    - from `AccountId32` to `AccountId` 
- and handling of deposit/withdraw for the sender's holding account.

mattsse:
## Changes

- Introduces the `MultiAssetAdapter` struct that acts as a configuration type to implement `TransactAsset`
- See also https://github.com/paritytech/polkadot/blob/master/xcm/xcm-builder/src/currency_adapter.rs#L46-L87
-

## Tests

```
cargo test
```

## Issues

- Closes #51


clearloop (approved):
LGTM, but no tests provide?

mattsse:
Not much to test for now, but it will be required when configuring XmcConfig.

PR #63, opened 2021-04-29
feat(committee): add constituent committee council selection

### LINKED ISSUE #4: Committee Pallet
The completion of this Epic requires the `Committee` pallet to be completed with intended features and tested.

## Implementation details
The PINT Committee utilizes the Committee pallet to manage on-chain proposals. 

## Acceptance Criteria
The completed pallet should have the following functionality:
- [x] `propose`
- [x] `vote`
- [x] `close`

### LINKED ISSUE #60: Add Constituent Committee council selection

clearloop:
* [PINT Governance](https://docs.polkadotindex.com/governance)

## Changes

This PR contains a function `propose_constituent` with an embedded storage rewriting function `add_constituent_encoded`, the process is like:

```
-> proposal_constituent(origin: council_member) 
-> propose(set_storage(new_committee_with_the_proposed_constituent))
-> voting
-> close with executing
```

## Tests

18e6fb8

## Issues

Setting storage directly for adding new constituent in the proposal, it would be better if constructing a call with root origin and wrapping it as an Action in `fn propose`

---

Closes #4 
Closes #60

[0]:https://docs.polkadotindex.com/pint-implementation-design#terminology


pallets/committee/src/lib.rs
```diff
@@ -380,6 +389,42 @@ pub mod pallet {
...
+        #[pallet::weight(10_000)] // TODO: Set weights
+        /// Propose new constituents to the constituent committee
+        ///
+        /// * This call only can be called by `MemberType::Council`
+        /// * This call will propose the call `set_members`
+        pub fn propose_constituents(
+            origin: OriginFor<T>,
+            _constituents: Vec<AccountIdFor<T>>,
+        ) -> DispatchResultWithPostInfo {
+            let _selector = Self::ensure_member(origin);
```
mattsse, line 412: I believe the decision to only allow council members to nominate new members is correct.
But `ensure_member` also succeeds for `Member::Constituent`, so we need to either `ensure!` that the member type is `Council` or add another ensure_method for `ensure_council`.

pallets/committee/src/lib.rs
```diff
@@ -380,6 +389,42 @@ pub mod pallet {
...
+        }
+
+        #[pallet::weight(10_000)] // TODO: Set weights
+        /// Propose new constituents to the constituent committee
+        ///
+        /// * This call only can be called by `MemberType::Council`
+        /// * This call will propose the call `set_members`
+        pub fn propose_constituents(
+            origin: OriginFor<T>,
+            _constituents: Vec<AccountIdFor<T>>,
```
mattsse, line 410: i doubt we will have several constituent members up for election at the same time, so this should be just a single value instead of an array.

pallets/committee/src/lib.rs
```diff
@@ -380,6 +389,42 @@ pub mod pallet {
...
+        ///
+        /// For the constituent selection momentary
+        pub fn execute(
+            origin: OriginFor<T>,
+            _proposal_hash: HashFor<T>,
+            // #[compact] length_bound: u32,
+        ) -> DispatchResultWithPostInfo {
+            Self::ensure_member(origin)?;
+            Ok(().into())
+        }
```
mattsse, line 427: Don't think we need another execute function, because `close` does already execute a successful proposal.

pallets/committee/src/lib.rs
```diff
@@ -380,6 +389,42 @@ pub mod pallet {
 
             Ok(().into())
         }
+
+        #[pallet::weight(10_000)] // TODO: Set weights
+        /// Set new constituent members
+        ///
+        /// This call can only be triggered by `fn propose_constituents` momentary
+        /// after approving of `propose_constituents`
+        pub fn set_members(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
mattsse, line 398:
```suggestion
        pub fn add_constituent(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```

mattsse, line 398: This will be the action of the constituent nominate proposal right?
Then this should ensure that the origin is approved by comittee first and then insert the new member.

pallets/committee/src/lib.rs
```diff
@@ -380,6 +389,42 @@ pub mod pallet {
...
+        /// Propose new constituents to the constituent committee
+        ///
+        /// * This call only can be called by `MemberType::Council`
+        /// * This call will propose the call `set_members`
+        pub fn propose_constituents(
+            origin: OriginFor<T>,
+            _constituents: Vec<AccountIdFor<T>>,
+        ) -> DispatchResultWithPostInfo {
+            let _selector = Self::ensure_member(origin);
+            Ok(().into())
```
mattsse, line 413: I think we can leverage the existing `propose` extrinsic and simply call this here with `::add_constituent` extrinsic as the action.

mattsse (changes requested):
The added extrinsics should the trick, what's missing is leveraging the existing `propose` extrinsic to connect them. The pipeline will probably look like `nominate_constituent` -> `propose(, add_constituent)`, does that make sense?

Cargo.toml
```diff
@@ -7,3 +7,6 @@ members = [
     'pallets/*',
     'runtime',
 ]
+
+[patch.crates-io]
+rocksdb = { git = "https://github.com/hdevalence/rust-rocksdb.git" }
```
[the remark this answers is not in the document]
clearloop: This is a patch on my machine temporally

clearloop:
This PR is not ready for review yet

pallets/committee/src/lib.rs
clearloop: Agree with a single value instead of an array

pallets/committee/src/lib.rs
clearloop: origin check is in the next commits, `ensure_member` for abstracting functions

pallets/committee/src/lib.rs
clearloop: removed

pallets/committee/src/lib.rs
clearloop: Wrapped this as a private KV set function instead a `Call`

pallets/committee/src/lib.rs
clearloop: The problem is it can't be wrapped into an `Action` directly, see https://github.com/ChainSafe/PINT/pull/63/commits/f3fde30c5cb27fb44f2a8745eb30f822d4a5ae92

pallets/committee/src/tests.rs
```diff
@@ -468,3 +468,43 @@ fn cannot_execute_proposal_twice() {
...
+
+            // close proposal
+            run_to_block(START_OF_V1 + 1);
+            assert_ok!(Committee::close(Origin::signed(EXECUTER_ACCOUNT_ID), hash,));
+        } else {
+            panic!("Could not get proposal hash from events");
+        }
+
+        // vote Aye on adding new constituent
+        <pallet::Members<Test>>::contains_key(42);
```
clearloop, line 508: Forgot to assert

pallets/committee/src/lib.rs
```diff
@@ -216,9 +230,13 @@ pub mod pallet {
         /// the committee
         pub fn ensure_member(
             origin: OriginFor<T>,
+            council: bool,
         ) -> Result<CommitteeMember<AccountIdFor<T>>, DispatchError> {
             let who = ensure_signed(origin)?;
             if let Some(member_type) = Members::<T>::get(who.clone()) {
+                let _ = council
+                    && member_type == MemberType::Constituent
+                    && Err(<Error<T>>::NotCouncilMember)?;
```
mattsse, line 239: This is rather hard to read. Should me make two designated functions for `ensure_council` and `ensure_consituent`instead?

pallets/committee/src/lib.rs
```diff
@@ -380,6 +425,34 @@ pub mod pallet {
...
+                Box::new(
+                    SystemCall::set_storage(Vec::from([Self::add_constituent_encoded(
+                        constituent,
+                    )?]))
+                    .into(),
+                ),
+            )?;
+
+            Ok(().into())
+        }
```
mattsse, line 455: After thinking about it again, I think we got the chain of events wrong. You're right we can't create an `Action` type on the fly because this type is configured in the runtime. So this will not be necessary anymore, instead we only require an `add_constituent` extrinsic.

pallets/committee/src/lib.rs
```diff
@@ -260,6 +278,33 @@ pub mod pallet {
                 })
             })
         }
+
+        /// Add a new constituent member and returns the storage key value pair
+        ///
+        /// This call only triggers after the approving of `propose_constituent`
+        fn add_constituent_encoded(
```
mattsse, line 285: This should be an extrinsic that ensures that it has been dispatched by the committee's origin, and if so adds a new constituent.

mattsse (changes requested):
you're right, there are limitations how we can create an `Action` type. We can't create it within dispatchable, because the type isn't known yet, it's configured in the runtime.
Looking at the [collective pallet](https://github.com/paritytech/substrate/blob/master/frame/collective/src/lib.rs) I think here is what we should do instead:
remove the `propose_constituent`, because we can't submit proposals from within this dispatchable.
Make the `add_constituent` function an extrinsic, so that we can submit this call as a proposal instead.
Now we need a way to verify within `add_constituent` that the origin is our `CommitteeOrigin::ApprovedByCommittee` type. This is probably related to #64

pallets/committee/src/lib.rs
clearloop: I like the current implementation as I provided, but I'm okay with changing it to `ensure_council` and `ensure_consituent`

pallets/committee/src/lib.rs
clearloop: Considered both an `extrinsic` and a `call` before, this `propose_constituent` is for the UX, a sugar for simplifying this action.

But due to the origin issue, I agree with changing it to an extrinsic now.

pallets/committee/src/types.rs
```diff
@@ -172,3 +175,32 @@ pub enum Vote {
...
+);
+impl<
+        O: Into<Result<CommitteeOrigin<AccountId, BlockNumber>, O>>
+            + From<CommitteeOrigin<AccountId, BlockNumber>>,
+        N: U32,
+        AccountId: PartialEq + Default,
+        BlockNumber: Default,
+    > EnsureOrigin<O> for EnsureApprovedByCommittee<N, AccountId, BlockNumber>
+{
+    type Success = ();
```
willemolding, line 194: Could make `type Success = AccountId` and then return the AccountId on ok. This would make it more consistent with the existing `ensure_signed`.

willemolding (approved):
This approach looks to me like it should work. One part I am not sure about is actually what the purpose is of having the `InitializeMembers` and `ChangeMembers` trait implementations at all if they are only called within the pallet anyway. @ansermino might be able to answer that.

Other than that looks good

clearloop:
> This approach looks to me like it should work. One part I am not sure about is actually what the purpose is of having the `InitializeMembers` and `ChangeMembers` trait implementations at all if they are only called within the pallet anyway. @ansermino might be able to answer that.
> 
> Other than that looks good

Long time no see! Willem!

mattsse (changes requested):
> One part I am not sure about is actually what the purpose is of having the InitializeMembers and ChangeMembers trait implementations at all if they are only called within the pallet anyway.

I also think we should consider removing them and instead adding a genesis config for setting the council in the genesis block.

pallets/committee/src/types.rs
clearloop: > Because the proposal is either accepted or rejected, don't think we consider cases were the vote count matters if a simple majority approved of the proposal?

This `U32` is for `min_votes`, https://github.com/ChainSafe/PINT/blob/604ff955fc3b282cba5801c172f9ff6af61aaf2d/pallets/committee/src/types.rs#L157-L160

mattsse: Yes, but that's already checked when the proposal is closed/executed 
https://github.com/ChainSafe/PINT/blob/604ff955fc3b282cba5801c172f9ff6af61aaf2d/pallets/committee/src/lib.rs#L361-L363

pallets/committee/Cargo.toml
```diff
@@ -12,10 +12,12 @@ version = '0.0.1'
 default = ['std']
 std = [
     'codec/std',
+    'serde/std',
```
mattsse, line 15:
```suggestion
    'serde',
```
is sufficient here

pallets/committee/src/lib.rs
```diff
@@ -115,6 +126,29 @@ pub mod pallet {
         OptionQuery,
     >;
 
+    #[pallet::genesis_config]
+    pub struct GenesisConfig<T: Config> {
+        pub members: Vec<T::AccountId>,
```
mattsse, line 131:
```suggestion
        pub council_members: Vec<T::AccountId>,
```
Maybe we should also make it possible to set `constituent` members in genesis?

pallets/committee/src/lib.rs
```diff
@@ -380,48 +417,23 @@ pub mod pallet {
...
-                    }
-                });
-            }
-            for leaving_member in outgoing {
-                Members::<T>::remove(leaving_member)
-            }
-            for member in incoming {
-                Members::<T>::insert(member, MemberType::Constituent);
-            }
+            Members::<T>::insert(constituent.clone(), MemberType::Constituent);
```
mattsse, line 433: We probably should we consider treating a `constituent` that is already a member, either council or constituent an error? Specifically the former case will result in a council member being stripped of their council privileges. This way it would be possible to demote a council member to constituent.

pallets/committee/src/mock.rs
```diff
@@ -107,9 +116,27 @@ pub fn run_to_block(n: u64) {
 }
 
 // Build genesis storage according to the mock runtime.
-pub fn new_test_ext() -> sp_io::TestExternalities {
-    let t = frame_system::GenesisConfig::default()
+pub fn new_test_ext<I>(members: I) -> sp_io::TestExternalities
```
mattsse, line 119:
```suggestion
pub fn new_test_ext(members: Vec<AccountId>) -> sp_io::TestExternalities
```
make this a vec instead?

mattsse (changes requested):
This looks good! Besides minor nits, I think we should check for duplicates when inserting a new constituent.

pallets/committee/src/mock.rs
clearloop: `IntoIterator<T>` includes `Vec` and the arguments could be shorter, for example

```rust
(PROPOSER_ACCOUNT_ID..PROPOSER_ACCOUNT_ID + 4)
// or
(
    PROPOSER_ACCOUNT_ID,
    PROPOSER_ACCOUNT_ID + 1,
    PROPOSER_ACCOUNT_ID + 2,
    PROPOSER_ACCOUNT_ID + 3,
)
```

pallets/committee/src/lib.rs
```diff
@@ -140,6 +183,10 @@ pub mod pallet {
         DuplicateVote,
         /// Attempted to cast a vote outside the accepted voting period for a proposal
         NotInVotingPeriod,
+        /// Attempted to add a member which is arealdy in council
+        ExistedCouncilMember,
```
mattsse, line 187:
```suggestion
        AlreadyCouncilMember,
```

pallets/committee/src/lib.rs
```diff
@@ -140,6 +183,10 @@ pub mod pallet {
...
+        /// Attempted to add a member which is arealdy a constituent
+        ExistedConstituentMember,
```
mattsse, line 189:
```suggestion
        AlreadyConstituentMember,
```

pallets/committee/src/lib.rs
```diff
@@ -140,6 +183,10 @@ pub mod pallet {
         DuplicateVote,
         /// Attempted to cast a vote outside the accepted voting period for a proposal
         NotInVotingPeriod,
+        /// Attempted to add a member which is arealdy in council
```
mattsse, line 186:
```suggestion
        /// Attempted to add a constituent that is already a member of the council
```

pallets/committee/src/lib.rs
```diff
@@ -140,6 +183,10 @@ pub mod pallet {
...
+        ExistedCouncilMember,
+        /// Attempted to add a member which is arealdy a constituent
```
mattsse, line 188:
```suggestion
        /// Attempted to add a constituent that is already a constituent
```

pallets/committee/src/lib.rs
```diff
@@ -380,48 +427,31 @@ pub mod pallet {
...
+
+            if let Some(ty) = <Members<T>>::get(constituent.clone()) {
+                return Err(match ty {
+                    MemberType::Council => <Error<T>>::ExistedCouncilMember,
+                    MemberType::Constituent => <Error<T>>::ExistedConstituentMember,
                 }
+                .into());
+            } else {
+                Members::<T>::insert(constituent.clone(), MemberType::Constituent);
             }
```
mattsse, line 451: We can use `try_mutate` here; if no value exist yet, we can simply replace the option, otherwise we throw based on the member (council, constituent)

mattsse (approved):
Minor nits and an optimization when inserting the member.

pallets/committee/src/tests.rs
```diff
@@ -468,3 +446,78 @@ fn cannot_execute_proposal_twice() {
...
+            approved_by_committee().into(),
+            42
+        ));
+
+        assert_noop!(
+            Committee::add_constituent(approved_by_committee().into(), 42),
+            <pallet::Error<Test>>::AlreadyConstituentMember
+        );
+    });
+}
```
clearloop, line 488: new tests

PR #68, opened 2021-04-30
feat(config): use statemint as parachain

mattsse:
## Changes

- Add statemint to polkadot-launch config to launch statemint as a parachain and create channel from pint to statemint
-
-

## Tests

```

```

## Issues

-

### DIFF
--- a/config.json
+++ b/config.json
@@ -22,25 +22,27 @@
   },
   "parachains": [
     {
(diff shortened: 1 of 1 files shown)


clearloop (approved):
Considering how could testing the integration in CI or Docker

Issue #32, opened 2021-04-07
RemoteTreasury pallet


dutterbutter:
## Implementation details

Similar to LocalTreasury with extra functionality to control assets on other chains via XCMP. The RemoteTreasury will be responsible for DOT, which will be controlled through XCMP on the relay chain.

## Acceptance Criteria

The completed pallet should have the following functionality:

- [x] `execute`

dutterbutter:
Closing, no longer required.

PR #69, opened 2021-05-11
feat(xcm): add xcm asset handler

mattsse:
## Changes

- Adds support type for executing cross chain messages
- intended to be used within the `RemoteAssetManager` and `RemoteTreasury`

## Tests

follow once integrated with #31 #32

## Issues

- Related #31 #32


clearloop (approved):
LGTM~

And maybe we should open an issue tracking the solutions of testing XCM transactions? 

...And why not configuring this pallet into runtime?

mattsse:
Created #70 to follow up on this

> ...And why not configuring this pallet into runtime?

This is not a pallet and will be configured in the runtime mod either with #31 or #32, whatever comes first.

PR #73, opened 2021-05-13
feat(asset-index): implement `remove_asset` API call

### LINKED ISSUE #24: Implement `remove_asset` API call
Implement `remove_asset` API call detailed in the specification under AssetIndex Pallet.
## Implementation details

- Dispatches transfer to move assets out of the index’s account, if a liquid asset is specified
- Updates the index to reflect the removed assets

|   Param   	| Type              	| Desc                                             	|
|:---------:	|-------------------	|--------------------------------------------------	|
| id        	| AssetId           	| Unique identifier of the asset being added       	|
| units     	| Balance           	| Amount of asset being added                      	|
| recipient 	| Option<AccountId> 	| If removing liquid assets, the recipient of them 	|

clearloop:
The [design](https://docs.polkadotindex.com/pint-implementation-design#assetindex) of `remove_asset`

## Changes

- Refactor trait `AssetRecorder`
- Implement `remove_asset` API Call

## Tests

```
$ cargo test
```

## Issues

Closes #24


pallets/asset-index/src/lib.rs
```diff
@@ -165,6 +176,56 @@ pub mod pallet {
...
+                        ExistenceRequirement::KeepAlive,
+                    )?;
+                } else {
+                    return Err(<Error<T>>::NoRecipient.into());
+                }
+            } else {
+                // TODO:
+                //
+                // Use `slash` or `resolving_into_exists` ?
+                T::IndexToken::slash(&caller, value);
```
clearloop, line 216: And maybe we should burn the value after this operation

clearloop, line 216: And maybe we should burn the value after this operation

pallets/asset-index/src/lib.rs
```diff
@@ -327,8 +388,47 @@ pub mod pallet {
...
-        fn remove_asset(_: &T::AssetId) -> DispatchResult {
-            todo!();
+        /// Remove asset with provied units.
+        ///
+        /// * If is liquid asset, transfer the ownership to the provied recipient
+        /// * Remove asset with provied units directly if is SAFT
+        fn remove_asset(
+            asset_id: &T::AssetId,
+            units: &T::Balance,
+            recipient: Option<MultiLocation>,
```
mattsse, line 522: I'm beginning to think that naming `remove_asset` in combination with the recipient argument is too ambiguous. Depending on the assetid to recipient combination, this call acts either as
* a transfer, if the asset is liquid and the recipient is present
* a burn otherwise

mattsse (commented):
I have to think about this more in detail

pallets/asset-index/src/lib.rs
```diff
@@ -165,6 +176,56 @@ pub mod pallet {
...
+        ///
+        /// Caller balance and recipient are updated to allocate
+        /// the correct amount of the IndexToken
+        pub fn remove_asset(
+            origin: OriginFor<T>,
+            asset_id: T::AssetId,
+            units: T::Balance,
+            recipient: Option<MultiLocation>,
+            recipient_account: Option<T::AccountId>,
+            value: T::Balance,
```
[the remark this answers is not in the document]
clearloop: TBH, I'm quite confused about the concepts `nav`, `PINT`, and `AssetIndex` LOL, I need to take more time to learn the relationship between them

pallets/asset-index/src/lib.rs
```diff
@@ -165,6 +176,56 @@ pub mod pallet {
...
+        /// Callable by an admin to remove assets from the index
+        /// and maybe transfer some IndexToken
+        ///
+        /// Caller balance and recipient are updated to allocate
+        /// the correct amount of the IndexToken
+        pub fn remove_asset(
+            origin: OriginFor<T>,
+            asset_id: T::AssetId,
+            units: T::Balance,
+            recipient: Option<MultiLocation>,
```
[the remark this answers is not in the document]
clearloop: If we don't redirect the `MultiLocation` in Liquid(MultiLocation), maybe we should burn the `liquid` assets and transfer `SAFT`?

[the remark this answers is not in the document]
clearloop: I'm thinking of a `liquid2SAFT` function, after removing liquid, we transfer SAFT to the recipient

liquid -> balance -> SAFT

Is this in design?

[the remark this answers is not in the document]
clearloop: BTW, should SAFT have an unique ASSET_ID?  If so, the `AssetAvailability` could be refactored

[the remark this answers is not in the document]
clearloop: And, if a price of an asset is dynamic, the balance recorded in TokenIndex should be dynamic as well

pallets/asset-index/src/lib.rs
```diff
@@ -182,6 +194,70 @@ pub mod pallet {
...
+                // If this value reach the total balance of IndexToken,
+                // modify to the max
+                let value = Self::calculate_pint_equivalent(asset_id.clone(), units)?;
+                let total_balance = T::IndexToken::total_balance(&caller);
+                if total_balance > value {
+                    value
+                } else {
+                    total_balance
+                }
+            };
```
clearloop, line 229: Need to check this

pallets/asset-index/src/lib.rs
```diff
@@ -130,6 +130,16 @@ pub mod pallet {
         /// A new asset was added to the index and some index token paid out
         /// \[AssetIndex, AssetUnits, IndexTokenRecipient, IndexTokenPayout\]
         AssetAdded(T::AssetId, T::Balance, AccountIdFor<T>, T::Balance),
+
+        // An asset was removed to the index and some index token transfered or slashed
```
mattsse, line 134:
```suggestion
        // An asset was removed from the index and some index token transfered or slashed
```

pallets/asset-index/src/lib.rs
```diff
@@ -130,6 +130,16 @@ pub mod pallet {
...
+        // \[AssetIndex, AssetUnits, IndexTokenRecipient, NewIndexTokenRecipient, IndexTokenPayout\]
```
mattsse, line 135:
```suggestion
        /// \[AssetIndex, AssetUnits, IndexTokenRecipient, NewIndexTokenRecipient, IndexTokenPayout\]
```

pallets/asset-index/src/lib.rs
```diff
@@ -450,8 +514,48 @@ pub mod pallet {
...
+        /// * If is liquid asset, transfer the ownership to the provied recipient
+        /// * Remove asset with provied units directly if is SAFT
+        fn remove_asset(
+            asset_id: &T::AssetId,
+            units: &T::Balance,
+            recipient: Option<MultiLocation>,
+        ) -> DispatchResult {
+            // TODO:
+            //
+            // Should we force check if AssetId and recipient are paired?
```
mattsse, line 528: can you elaborate on this?

pallets/asset-index/src/lib.rs
```diff
@@ -450,8 +514,48 @@ pub mod pallet {
...
+            //
+            // For example:
+            //
+            // If `!is_liquid`, the recipient must be none?
+            Holdings::<T>::try_mutate(asset_id, |value| -> Result<_, Error<T>> {
+                let mut index_asset_data =
+                    value.clone().ok_or(<Error<T>>::UnsupportedAsset.into())?;
```
mattsse, line 535:
```suggestion
                    value.take().ok_or(<Error<T>>::UnsupportedAsset.into())?;
```

pallets/asset-index/src/lib.rs
```diff
@@ -450,8 +514,48 @@ pub mod pallet {
...
+                if index_asset_data.is_liquid() {
+                    // If is liquid asset, transfer ownership to the provied
+                    // recipient or throw error
+                    //
+                    // TODO:
+                    //
+                    // XCM transfer?
```
mattsse, line 542: this we need to think about, whether `remove_asset` has any xcm side effect

pallets/asset-index/src/lib.rs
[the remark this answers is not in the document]
mattsse: all assets have an unique asset id. 
They way this is supposed to work, is that comittee votes on adding assets.
I think we should probably split this process into a separate `register_asset` that sets the loctation

pallets/asset-index/src/lib.rs
```diff
@@ -182,6 +196,56 @@ pub mod pallet {
...
+                        account,
+                        value,
+                        ExistenceRequirement::KeepAlive,
+                    )?;
+                } else {
+                    return Err(<Error<T>>::NoRecipient.into());
+                }
+            } else {
+                T::IndexToken::slash(&caller, value.clone());
+                T::IndexToken::burn(value);
```
mattsse, line 236: This is not ideal, because there is no guarantee that we have the account has enough balance, I think `burn` -> `settle` would be more idiomatic.

pallets/asset-index/src/lib.rs
```diff
@@ -182,6 +196,56 @@ pub mod pallet {
...
+                if let Some(account) = &recipient_account {
+                    T::IndexToken::transfer(
+                        &caller,
+                        account,
+                        value,
+                        ExistenceRequirement::KeepAlive,
+                    )?;
+                } else {
+                    return Err(<Error<T>>::NoRecipient.into());
+                }
```
mattsse, line 233: After reading the docs again, I believe the expected behaviour would be to send an xcm to the recipient's location, instead of an internal transfer

pallets/asset-index/src/lib.rs
```diff
@@ -182,6 +196,56 @@ pub mod pallet {
...
+            asset_id: T::AssetId,
+            units: T::Balance,
+            recipient: Option<MultiLocation>,
+            recipient_account: Option<T::AccountId>,
+        ) -> DispatchResultWithPostInfo {
+            T::AdminOrigin::ensure_origin(origin.clone())?;
+            let caller = ensure_signed(origin)?;
+            <Self as AssetRecorder<T::AssetId, T::Balance>>::remove_asset(
+                &asset_id, &units, recipient,
+            )?;
```
mattsse, line 216: This modifies while following operations are fallible, so we need to either mark this extrinsic make it mutation safe.

mattsse (changes requested):
This is certainly going in the right direction.
There is more work to be done to make this work as expected and safe.
Sorry for the confusion around the transfer part

pallets/asset-index/src/lib.rs
clearloop: For example, both `Some` or both `None`, this is because we only have a `remove_assset`

pallets/asset-index/src/lib.rs
clearloop: The `remove_asset` call only contains this one mutating storage operation, and it runs at the top of the call `remove_asset`

clearloop:
I'll back to this PR after #82

mattsse:
#135 made some changes wrt how assets are added. Now `remove_assets` check first if it's liquid or not.
* if liquid, we can use the `RemoteAssetManager::transfer` function to send to the asset's chain.

In both cases we need as parameters `asset_id` and `units` and need to burn or (T::Currency::withdraw) them from the index's account.
The additional parameter `recepient`  is mandatory if it's a liquid asset. so that we can actually transfer them.

Maybe it makes sense to split this into `remove_saft` and `remove_liquid`, not sure about the naming at this point

clearloop:
Move forward to a new PR

Issue #70, opened 2021-05-12
Add testing support for XCM


mattsse:
## Issue summary

- Tracking issue for how to test xcm properly

## Other information and links

- Related [shaunxw/xcm-simulator](https://github.com/shaunxw/xcm-simulator)

Related #31 #32

mattsse:
https://github.com/shaunxw/xcm-simulator was just updated and orml has some examples how to implement it

PR #77, opened 2021-05-14
feat(index): implement withdraw extrinsic

### LINKED ISSUE #26: Implement `withdraw` API call
Implement `withdraw` API call detailed in the specification under AssetIndex Pallet.
## Implementation details

- Queries PriceFeed for NAV of index token and calculates proportions of underlying assets to be withdrawn
- Dispatches calls to other chains to unbond underlying assets
- Updates index. If this reduces the total supply of any asset to 0, removes asset type from index entirely
- Burns an amount of the index token specified by its NAV
- Amount of index token must be above MinimumRedemption threshold

|  Param 	| Type    	| Desc                                      	|
|:------:	|---------	|-------------------------------------------	|
| amount 	| Balance 	| Amount of the index token being withdrawn 	|

mattsse:
## Changes

- Implements the `withdraw` extrinsic
- Only liquid assets are considered when calculating the distribution

## Tests

```
cargo test
```

## Issues

- Closes #26

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -28,13 +28,15 @@ pub mod pallet {
     use crate::types::{
         AssetAvailability, AssetWithdrawal, IndexAssetData, PendingRedemption, RedemptionState,
     };
+    use frame_support::sp_runtime::traits::{AccountIdConversion, Saturating};
     use frame_support::sp_runtime::{FixedPointNumber, FixedU128};
+    use frame_support::traits::{ExistenceRequirement, WithdrawReasons};
     use frame_support::{
         dispatch::DispatchResultWithPostInfo,
         pallet_prelude::*,
         sp_runtime::traits::{AtLeast32BitUnsigned, CheckedAdd, CheckedDiv, CheckedSub, Zero},
         sp_std::{convert::TryInto, prelude::*, result::Result},
-        traits::{Currency, LockableCurrency},
+        traits::{Currency, Imbalance, LockableCurrency},
         PalletId,
     };
     use frame_system::pallet_prelude::*;
@@ -122,12 +124,15 @@ pub mod pallet {
     #[pallet::metadata(T::AssetId = "AccountId", AccountIdFor<T> = "AccountId", T::Balance = "Balance")]
     #[pallet::generate_deposit(pub(super) fn deposit_event)]
     pub enum Event<T: Config> {
-        // A new asset was added to the index and some index token paid out
-        // \[AssetIndex, AssetUnits, IndexTokenRecipient, IndexTokenPayout\]
+        /// A new asset was added to the index and some index token paid out
+        /// \[AssetIndex, AssetUnits, IndexTokenRecipient, IndexTokenPayout\]
         AssetAdded(T::AssetId, T::Balance, AccountIdFor<T>, T::Balance),
-        // A new deposit of an asset into the index has been performed
-        // \[AssetId, AssetUnits, Account, PINTPayout\]
+        /// A new deposit of an asset into the index has been performed
+        /// \[AssetId, AssetUnits, Account, PINTPayout\]
         Deposited(T::AssetId, T::Balance, AccountIdFor<T>, T::Balance),
+        /// Started the withdrawal process
+        /// \[Account, PINTAmount\]
+        WithdrawalInitiated(AccountIdFor<T>, T::Balance),
     }
 
     #[pallet::error]
@@ -207,7 +212,7 @@ pub mod pallet {
             let issued = T::IndexToken::issue(pint_amount);
 
             // add minted PINT to user's balance
-            T::IndexToken::resolve_creating(&caller, pint_amount);
+            T::IndexToken::resolve_creating(&caller, issued);
             Self::deposit_event(Event::Deposited(asset_id, amount, caller, pint_amount));
             Ok(().into())
         }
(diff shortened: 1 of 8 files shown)


pallets/asset-index/src/lib.rs
```diff
@@ -25,22 +25,30 @@ pub mod pallet {
     use crate::traits::WithdrawalFee;
     pub use crate::traits::{AssetRecorder, MultiAssetRegistry};
     pub use crate::types::MultiAssetAdapter;
-    use crate::types::{AssetAvailability, IndexAssetData, PendingRedemption};
+    use crate::types::{
+        AssetAvailability, AssetWithdrawal, IndexAssetData, PendingRedemption, RedemptionState,
+    };
+    use frame_support::sp_runtime::traits::{AccountIdConversion, Saturating};
+    use frame_support::sp_runtime::{FixedPointNumber, FixedU128};
+    use frame_support::traits::{ExistenceRequirement, WithdrawReasons};
```
clearloop, line 33: Why don't we write like this

```rust
use frame_support::{
    sp_runtime::traits::{AccountIdConversion, Saturating},
    frame_support::sp_runtime::{FixedPointNumber, FixedU128},
    frame_support::traits::{ExistenceRequirement, WithdrawReasons},
}
```

pallets/asset-index/src/lib.rs
```diff
@@ -180,7 +195,7 @@ pub mod pallet {
                 .filter(|holding| matches!(holding.availability, AssetAvailability::Liquid(_)))
                 .ok_or(Error::<T>::UnsupportedAsset)?;
 
-            let pint_amount = Self::calculate_asset_nav(asset_id.clone(), amount)?;
+            let pint_amount = Self::calculate_pint_equivalent(asset_id.clone(), amount)?;
```
clearloop, line 198: This is more clear 👍

pallets/remote-asset-manager/src/lib.rs
```diff
@@ -98,5 +98,16 @@ pub mod pallet {
         ) -> DispatchResult {
             todo!()
         }
+
+        fn bond(_asset: <T as Config>::AssetId, _amount: <T as Config>::Balance) -> DispatchResult {
+            todo!()
```
clearloop, line 103: Guess these `todos!()`s make the CI sad

pallets/asset-index/src/lib.rs
```diff
@@ -215,19 +235,108 @@ pub mod pallet {
...
+            T::IndexToken::ensure_can_withdraw(
+                &caller,
+                amount,
+                WithdrawReasons::TRANSFER,
+                free_balance.saturating_sub(amount),
+            )?;
 
             let fee = T::WithdrawalFee::withdrawal_fee(amount);
-            let _redeem = amount
+            let redeem = amount
```
clearloop, line 247: I have a question:

```text
if an account wants to redeem 10 unit balance exactly && crazily,
how could (s)he make that?
```

I think maybe we should think more about this `sub(&fee)` directly in the amount? Or just leave this question to the client of PINT

pallets/asset-index/src/lib.rs
clearloop, line 198: This is more clear 👍

pallets/asset-index/src/lib.rs
clearloop, line 33: Why wouldn't we write like this?

```rust
use frame_support::{
    sp_runtime::traits::{AccountIdConversion, Saturating},
    frame_support::sp_runtime::{FixedPointNumber, FixedU128},
    frame_support::traits::{ExistenceRequirement, WithdrawReasons},
}
```

pallets/asset-index/src/lib.rs
```diff
@@ -215,19 +235,108 @@ pub mod pallet {
...
+            // update the index balance by burning all of the redeemed tokens and
+            // issuing new tokens for the tx fee into the treasury's account
+            let burned = T::IndexToken::burn(amount);
+            if let Err(burned) = T::IndexToken::settle(
+                &caller,
+                burned,
+                WithdrawReasons::TRANSFER,
+                ExistenceRequirement::KeepAlive,
+            ) {
+                // TODO can this even happen after ensure_can_withdraw?
```
clearloop, line 287: I think it would be too strict if we burn balances on the failing of `ensure_can_withdraw`, users just paid the tx fee.

To be confirmed, the design of the tx fee in this method exists along with the custom transaction fee calculated by the logic of substrate framework ( the fee `T::Weight`  determines )?

pallets/asset-index/src/lib.rs
```diff
@@ -215,19 +235,108 @@ pub mod pallet {
...
+
+                assets.push(AssetWithdrawal {
+                    asset,
+                    state,
+                    units,
+                });
+            }
+
+            // lock the assets for the withdrawal period starting at current block
+            PendingWithdrawals::<T>::mutate(&caller, |maybe_redemption| {
```
clearloop, line 330: Maybe we can add docs to our mutating operations to declare our mutations are safe when our function doesn't have a wrapper like `with_transaction_result`

clearloop (approved):
LGTM! 

Tests are required, linking #70

pallets/remote-asset-manager/src/lib.rs
mattsse: Makes sense for now, would also make testing easier

pallets/asset-index/src/lib.rs
mattsse: ideally the user would receive the distribution of the underlying (liquid) assets in proportion to their representation in the index.
Therefor we need to calculate the asset's proportion first and then its PINT equivalent:

```
NAV(amount) = NAV_asset_proportional_0 + NAV_asset_proportional_1 ... + NAV_asset_proportional_n - Fee
```

where `NAV_asset_proportional_n = NAV_asset_n / NAV_liquid * (amount- fee) * Price(PINT/n)`

pallets/asset-index/src/lib.rs
mattsse: that's a good point, I have to evaluate this.

mattsse:
@clearloop I've fixed a bug that resulted in the wrong asset ratio and accounted for rounding when calculating the distribution

Issue #74, opened 2021-05-13
Implement `transfer` API call


dutterbutter:
Implement `transfer` API call detailed in the specification under RemoteAssetManager Pallet.

### Implementation Details

Dispatch XCM message to transfer asset between caller account and recipient

|  Param 	| Type    	| Desc                                      	|
|:------:	|---------	|-------------------------------------------	|
| id 	| AssetId 	| Assets being transferred 	|
| units 	| Balance 	| Assets being transferred 	|
| recipient 	| AccountId 	| Recipient of transfer	|

PR #100, opened 2021-06-08
feat: add assets pallet xcm bindings

### LINKED ISSUE #99: Add support for transacting pallet_asset calls
## Issue summary

- The FRAME `asset` pallet provides extrinsics to manage tokens, In order to support `assets` pallet token on PINT, transacting calls is necessary 

- See also #98 
- [statemint](https://github.com/paritytech/statemint)

mattsse:
## Changes

- Adds bindings for the assets pallets

## Tests

```
cargo test
```

## Issues

- Closes #99 
- Ref #58 #74


mattsse:
Here is the expanded assets pallet
[assets_expanded.rs.txt](https://github.com/ChainSafe/PINT/files/6616826/assets_expanded.rs.txt)

clearloop (approved):
This part is from https://github.com/paritytech/substrate/blob/master/frame/assets/src/lib.rs? Cool, but why wouldn't we just import or wrap the pallets? it would be easier to pull the code instead of developing and updating ourselves

mattsse:
> This part is from https://github.com/paritytech/substrate/blob/master/frame/assets/src/lib.rs? Cool, but why wouldn't we just import or wrap the pallets? it would be easier to pull the code instead of developing and updating ourselves

I also thought about this, the challenge is all the (potentially) different associated types of a pallet of different chains. We could pull in all the different runtimes directly, however this has some drawbacks as well:
* we need to compile all dependent runtimes everytime
* supporting different chains is now only possible in the runtime itself, adding/removing one will require runtime upgrade
* getting conversion right on type level is rather difficult and you need something like this when configuring it: 

```rust
    type AssetsPallet: pallt_assets::Config< AccountId : From<T::Account> >
```

However the whole problem is that all the `pallet_assets::Config::<*Associated type*>` are potentially different for each chain so we need always different conversion from `PINT::<*Associated type*>` to `OtherChain::<*Associated type*`

However this may be overengineered and not really necessary, especially for the pallet_assets support, if we only targeting `statemint/mine`

clearloop:
> I also thought about this, the challenge is all the (potentially) different associated types of a pallet of different chains. We could pull in all the different runtimes directly, however, this has some drawbacks as well

I'm not sure if we have to do this, but I'm okay with this.

>  you mean implementing the pallets on a different type that we then plug into the asset manager as associated type

Yep, based on the existed pallets with customized associated types as implementations

PR #104, opened 2021-06-09
feat(committee): add CommitteeMember origin for pallet committee

### LINKED ISSUE #96: Set initial council members in runtime
## Issue summary

Note: The origin configuration of the pallet committee has been modified to `EnsureSigned` in #91, but we still don't have initial council members in the runtime configuration.

- #94 
- #64 
https://github.com/paritytech/substrate/blob/37bb3ae7eb559afaf9c7dbf7fd99e08b282c8127/bin/node/runtime/src/lib.rs#L494

clearloop:
## Changes

- [x] check votes in `EnsureAprovedByCommitte`
- [x] add `CommitteeMember` origin
- [x] add initial committee members in the genesis config
- [x] update tests
- [x] update benchmarks

## Tests

```
CI
```

## Issues

Closes #96

### DIFF
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -322,8 +322,8 @@ pub mod pallet {
         /// The provided action will be turned into a proposal and added to the list of current active proposals
         /// to be voted on in the next voting period.
         pub fn propose(origin: OriginFor<T>, action: Box<T::Action>) -> DispatchResultWithPostInfo {
-            let proposer = ensure_signed(origin.clone())?;
-            T::ProposalSubmissionOrigin::ensure_origin(origin)?;
+            T::ProposalSubmissionOrigin::ensure_origin(origin.clone())?;
+            let proposer = ensure_signed(origin)?;
 
             // Create a new proposal with a unique nonce
             let nonce = Self::take_and_increment_nonce()?;
@@ -390,8 +390,8 @@ pub mod pallet {
             origin: OriginFor<T>,
             proposal_hash: HashFor<T>,
         ) -> DispatchResultWithPostInfo {
-            let closer = ensure_signed(origin.clone())?;
-            T::ProposalExecutionOrigin::ensure_origin(origin)?;
+            T::ProposalExecutionOrigin::ensure_origin(origin.clone())?;
+            let closer = ensure_signed(origin)?;
 
             // ensure proposal has not already been executed
             ensure!(
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -87,7 +87,7 @@ pub mod pallet {
             Success = <Self as frame_system::Config>::AccountId,
         >;
 
-        /// Origin that is permitted to execute approved proposals
+        /// Origin that is permitted to vote
         type ProposalVoteOrigin: EnsureOrigin<
             <Self as frame_system::Config>::Origin,
             Success = <Self as frame_system::Config>::AccountId,
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -323,7 +323,7 @@ pub mod pallet {
         /// to be voted on in the next voting period.
         pub fn propose(origin: OriginFor<T>, action: Box<T::Action>) -> DispatchResultWithPostInfo {
             T::ProposalSubmissionOrigin::ensure_origin(origin.clone())?;
-            let proposer = ensure_signed(origin)?;
+            let proposer = Self::ensure_member(origin)?;
 
             // Create a new proposal with a unique nonce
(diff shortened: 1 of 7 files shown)


pallets/committee/src/lib.rs
```diff
@@ -76,10 +76,22 @@ pub mod pallet {
...
+        type ProposalExecutionOrigin: EnsureOrigin<
+            <Self as frame_system::Config>::Origin,
+            Success = <Self as frame_system::Config>::AccountId,
+        >;
+
+        /// Origin that is permitted to vote
+        type ProposalVoteOrigin: EnsureOrigin<
+            <Self as frame_system::Config>::Origin,
+            Success = <Self as frame_system::Config>::AccountId,
+        >;
```
mattsse, line 94: what's the reasoning behind this extra origin? Aren't only member's allowed to vote?

pallets/committee/src/lib.rs
```diff
@@ -357,7 +355,11 @@ pub mod pallet {
             vote: Vote,
         ) -> DispatchResultWithPostInfo {
             // Only members can vote
-            let voter = Self::ensure_member(origin)?;
+            let caller = T::ProposalVoteOrigin::ensure_origin(origin)?;
+            let voter = CommitteeMember::new(
+                caller.clone(),
+                <Members<T>>::get(caller).ok_or(<Error<T>>::NotMember)?,
+            );
```
mattsse, line 362: Since only members are allowed to vote, I believe we should only check for inclusion in `Members` instead adding another origin, which will probably do the same as `ensure_member`

pallets/committee/src/types.rs
```diff
@@ -208,3 +214,34 @@ impl<
...
+            + Clone,
+        T: Config,
+    > EnsureOrigin<O> for EnsureMember<T>
+{
+    type Success = <T as frame_system::Config>::AccountId;
+    fn try_origin(o: O) -> Result<Self::Success, O> {
+        let origin = o.clone().into()?;
+        match origin {
+            RawOrigin::Signed(i) => {
+                if <Members<T>>::contains_key(i.clone()) {
```
mattsse, line 233: does `contains_key` require ownership of `i`, hence the clone?

mattsse (changes requested):
The Origin types look great. 
But I don't think we need it necessarily need it in the `Committee` pallet as extra config type, since we can check for inclusion in the `Members` directly. Because we check for inclusion anyway

pallets/committee/src/lib.rs
clearloop: ur right LOL, added an origin for Vote creates a lot of trouble for me QAQ

But the logic of `EnsureMember` contains the logic of `ensure_member`, so I want to remove `ensure_member` crazily LOL

pallets/committee/src/types.rs
clearloop: Oops, this is a nit caused the previous implementation

pallets/committee/src/lib.rs
mattsse: > BTW, I'm not afraid of Origin anymore now hhh

That's awesome !

clearloop:
> But I don't think we need it necessarily need it in the Committee pallet as extra config type, since we can check for inclusion in the Members directly. Because we check for inclusion anyway

This is also a confusing thing while I working on this PR, the difference between our `pallet-committee` and `pallet-collective` of substrate, we don't need to call `propose`, `vote`, `close` outside `pallet-committee`, so the `ProposalSubmissionOrigin` and `ProposalExecutionOrigin` might also not necessary

But for pallets like `asset-index`, add or remove asset, we might need the `EnsureMember` provided by this pallet.

pallets/committee/src/mock.rs
```diff
@@ -88,15 +88,15 @@ ord_parameter_types! {
...
+    crate::EnsureApprovedByCommittee<Test>,
 >;
 
 impl pallet_committee::Config for Test {
     type ProposalSubmissionPeriod = ProposalSubmissionPeriod;
     type VotingPeriod = VotingPeriod;
     type MinCouncilVotes = MinCouncilVotes;
-    type ProposalSubmissionOrigin = frame_system::EnsureSignedBy<AdminAccountId, AccountId>;
-    type ProposalExecutionOrigin = frame_system::EnsureSignedBy<ExecuterAccountId, AccountId>;
+    type ProposalSubmissionOrigin = EnsureMember<Self>;
```
mattsse, line 98: This should be `EnsureSigned` only since, anyone is allowed to submit proposals

mattsse (approved):
It's intended that anyone can submit a proposal, so this should just be the standard EnsureSigned

You're probably right, however this may change, so my feeling is we should keep the origins in

PR #105, opened 2021-06-10
feat(asset-index): add AssetMetadata

### LINKED ISSUE #102: Refactor `AssetId` in pallet-asset-index as enumeration
## Issue summary

> one question about the From<u32> restriction, was this necessary for the benchmarks?
> I don't think we should necessarily restrict it to be an u32, for example it could be end up something like Acala's currencyId
>
> https://github.com/AcalaNetwork/Acala/blob/master/primitives/src/currency.rs#L208-L212

-  https://github.com/ChainSafe/PINT/pull/91#pullrequestreview-678330525
- #78

### LINKED ISSUE #115: Refactor AssetId
## Issue summary

Currently, we have numeric AssetId in `pallet-asset-index`, should we refactor it into another type?

https://github.com/ChainSafe/PINT/blob/49ce26841414367682fca48f25874ac55c6822a6/pallets/asset-index/src/lib.rs#L94

- [Acala/primitives/Currency](https://github.com/AcalaNetwork/Acala/blob/4dedc2feaddf416861c26af861edb31716f7ac62/primitives/src/currency.rs#L208-L212)
- [polkadot/xcm/xcm-executor](https://github.com/paritytech/polkadot/blob/c1940f13aeebf1e46be6f65fe14c0cbd65af8c77/xcm/xcm-executor/src/assets.rs#L21)
- https://github.com/AcalaNetwork/Acala/pull/1073

- #105
- #102

clearloop:
After thinking about it over again, I don't think refactoring `AssetId` to `enum` is a better idea, the following is what `AssetId` is in my current understanding

* AssetId is unique for PINT assets
  * But one asset can have different AssetIds since Assets could be submitted by different accounts.
  * So if we use `enum`, the `enum` have to wrap something which can make Assets different from others, and the easiest and the most practical type is `number`
* The assets PINT served are dynamic
  * Once an asset has been removed, the match AssetId should not be used anymore
  * Whatever any types we set for Assetid, it's hard to avoid using `number`

---

* [Acala/primitives/Currency](https://github.com/AcalaNetwork/Acala/blob/4dedc2feaddf416861c26af861edb31716f7ac62/primitives/src/currency.rs#L208-L212)
* [polkadot/xcm/xcm-executor](https://github.com/paritytech/polkadot/blob/c1940f13aeebf1e46be6f65fe14c0cbd65af8c77/xcm/xcm-executor/src/assets.rs#L21)

## Changes

- `AssetId: + AtLeast32BitUnsigned`

## Tests

```
CI
```

## Issues

Closes #102 
Closes #115


mattsse:
> AssetId is unique for PINT assets
> * But one asset can have different AssetIds since Assets could be submitted by different accounts.

This is one not clear for me, Assets can be submitted by different accounts, but are always unique.

> The assets PINT served are dynamic
> * Once an asset has been removed, the match AssetId should not be used anymore
> * Whatever any types we set for Assetid, it's hard to avoid using number

This is a very good point.

Yes unique int ids make the most sense, I guess the `assets` pallet does the same, did you have a look?

What we certainly need is some kind of additional metadata struct for every asset, something like

```
struct AssetInfo<Id> {
    name : Vec<u8>,
    symbol: Vec<u8>,
    id: Id
}
```
similar to chainlink feed info

I don't think we have a proper storage map for this, we have the `MultiAssetRegistry` trait that is implemented for the `asset_index`, that should be extended

pallets/asset-index/src/types.rs
```diff
@@ -33,16 +33,36 @@ pub enum AssetAvailability {
     Saft,
 }
 
+/// Metadata for an asset
+#[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug)]
+pub struct AssetMetadata {
+    pub name: Vec<u8>,
+    pub symbol: Vec<u8>,
+    pub decimals: u8,
+}
```
clearloop, line 42: Use fixed `Vec<u8>` or associated types?

mattsse: We need something like this https://github.com/paritytech/substrate/blob/master/frame/assets/src/lib.rs#L264 that configures the metdata https://github.com/paritytech/substrate/blob/master/frame/assets/src/types.rs#L92

so we need a `StringLimit` type and make these vecs bounded

pallets/asset-index/src/lib.rs
```diff
@@ -525,13 +528,18 @@ pub mod pallet {
         /// Creates IndexAssetData if entry with given assetID does not exist.
         /// Otherwise adds the units to the existing holding
         fn add_asset(
+            metadata: AssetMetadata,
             asset_id: &T::AssetId,
```
mattsse, line 532: This means `metadata` is required for every new `add_asset` call?
Can we extract this into a separate functions `set_metadata`?

pallets/asset-index/src/lib.rs
```diff
@@ -525,13 +528,18 @@ pub mod pallet {
         /// Creates IndexAssetData if entry with given assetID does not exist.
         /// Otherwise adds the units to the existing holding
         fn add_asset(
+            metadata: AssetMetadata,
```
mattsse, line 531: This means metadata is required for every new add_asset call?
Can we extract this into a separate functions like `set_metadata`?

pallets/asset-index/src/lib.rs
```diff
@@ -525,13 +528,18 @@ pub mod pallet {
...
             asset_id: &T::AssetId,
             units: &T::Balance,
             availability: &AssetAvailability,
         ) -> DispatchResult {
             Holdings::<T>::try_mutate(asset_id, |value| -> Result<_, Error<T>> {
                 let index_asset_data = value.get_or_insert_with(|| {
-                    IndexAssetData::<T::Balance>::new(T::Balance::zero(), availability.clone())
+                    IndexAssetData::<T::Balance>::new(
+                        metadata,
```
mattsse, line 539: I think we can keep this separate, maybe introduce a designated storage map for metadata, like: https://github.com/paritytech/substrate/blob/master/frame/assets/src/lib.rs#L260

mattsse (changes requested):
I think keeping `Metadata` separate will be more convenient

pallets/asset-index/src/lib.rs
clearloop: > Can we extract this into a separate functions like set_metadata?

Not sure about this, for example, if we want to issue a token on ethereum, we set the metadata when we issue it

mattsse: This is different, we can add/modify the asset's metadata on the fly, we also don't need  actually require it.

runtime/src/lib.rs
```diff
@@ -617,6 +617,7 @@ parameter_types! {
     pub MinimumRedemption: u32 = 0;
     pub WithdrawalPeriod: <Runtime as frame_system::Config>::BlockNumber = 10;
     pub DOTContributionLimit: Balance = 999;
+    pub PalletIndexStringLimit: u32 = 8;
```
mattsse, line 620: This might be too low, bumping this to 50 should be fine
```suggestion
    pub PalletIndexStringLimit: u32 = 50;
```

mattsse (approved):
LGTM

mattsse (approved):

PR #106, opened 2021-06-10
feat(asset-index): store initial price pair

### LINKED ISSUE #93: `AssetPriceFeed not found` in the call `deposit` of pallet `asset-index`
## Issue summary

This `deposit`  function runs well in tests, but not in benchmarks, besides, met this issue also while I was writing the Javascript API of PINT in #83.

Guess need to ensure that a valid round has already submitted first, separating from #91 

- https://github.com/ChainSafe/PINT/blob/4fcfc4fda293c1a973dd97ac8e1dd7010e4d944d/pallets/asset-index/src/tests.rs#L91
- https://github.com/ChainSafe/PINT/blob/4fcfc4fda293c1a973dd97ac8e1dd7010e4d944d/pallets/price-feed/src/lib.rs#L175

---

Could not use `pallet_chainlink_feed` in the benchmark of `pallet_asset_index`, these two pallets are not tight coupling, #94

### LINKED ISSUE #94: Check if feed pair exists while adding asset in pallet asset index
## Issue summary

The call `add_asset` in pallet `asset-index` adds assets directly without checking if the matched feed pair exists which make the following calls like `deposit` could not execute

- #93 
- https://github.com/ChainSafe/PINT/blob/4fcfc4fda293c1a973dd97ac8e1dd7010e4d944d/pallets/price-feed/src/tests.rs#L90

### LINKED ISSUE #95: Store initial price pair
## Issue summary

- as discussed in #94 : `add_asset` determines PINT/asset and should be stored as initial price for this asset
- this way it's guaranteed that at least a single PINT/asset pair is available

-

clearloop:
## Changes

- [x] Add checks for dup assets
- [x] store initial price pair
- [x] tests
- [ ] benchmarks

## Tests

```
CI
```

## Issues

Ref #105 
Closes #93 
Closes #94 
Closes #95

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -42,7 +42,7 @@ pub mod pallet {
     use xcm::opaque::v0::MultiLocation;
 
     use pallet_asset_depository::MultiAssetDepository;
-    use pallet_price_feed::{AssetPricePair, PriceFeed};
+    use pallet_price_feed::{AssetPricePair, Price, PriceFeed};
     use pallet_remote_asset_manager::RemoteAssetManager;
 
     use crate::traits::WithdrawalFee;
@@ -192,6 +192,10 @@ pub mod pallet {
         ) -> DispatchResultWithPostInfo {
             T::AdminOrigin::ensure_origin(origin.clone())?;
             let caller = ensure_signed(origin)?;
+
+            // Store intial price pair if not exists
+            T::PriceFeed::ensure_price(asset_id, Price::from_inner((value / units).into()))?;
+
             <Self as AssetRecorder<T::AssetId, T::Balance>>::add_asset(
                 &asset_id,
                 &units,
@@ -524,12 +528,6 @@ pub mod pallet {
             units: &T::Balance,
             availability: &AssetAvailability,
         ) -> DispatchResult {
-            if T::PriceFeed::get_price(*asset_id).is_err() {
-                // init price feed if not exists
-            } else {
-                return Err(<Error<T>>::AssetExisted.into());
-            }
-
             Holdings::<T>::try_mutate(asset_id, |value| -> Result<_, Error<T>> {
                 let index_asset_data = value.get_or_insert_with(|| {
                     IndexAssetData::<T::Balance>::new(T::Balance::zero(), availability.clone())
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -158,6 +158,8 @@ pub mod pallet {
     pub enum Error<T> {
         /// Thrown if adding units to an asset holding causes its numerical type to overflow
         AssetUnitsOverflow,
+        /// Thrown if adding an existed asset
+        AssetExisted,
         /// Thrown if no index could be found for an asset identifier.
         UnsupportedAsset,
         /// Thrown if calculating the volume of units of an asset with it's price overflows.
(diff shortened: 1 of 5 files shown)


pallets/price-feed/src/lib.rs
```diff
@@ -76,6 +76,16 @@ pub mod pallet {
...
         StorageMap<_, Blake2_128Concat, T::AssetId, FeedIdFor<T>, OptionQuery>;
 
+    #[pallet::storage]
+    /// (AssetId) -> AssetPricePair
+    ///
+    /// This storage stores the initial price pair for quote assets based on `SelfAssetId`
+    ///
+    /// * insert: adding a new asset with no price pair been set yet
+    /// * remove: an asset has its price pair, remove automatically
+    pub type InitialPricePair<T: Config> =
```
clearloop, line 86: Use an `InitialPricePair` but not submit new feeds

pallets/price-feed/src/lib.rs
```diff
@@ -272,6 +282,30 @@ pub mod pallet {
 
             Ok(AssetPricePair { base, quote, price })
         }
+
+        /// Ensure quote asset has price pair based on `Self::AssetId`
```
mattsse, line 286:
```suggestion
        /// Ensure quote asset has a price pair based on `Self::AssetId`
```

mattsse, line 286:
```suggestion
        /// Ensure quote asset has an initial price pair
```

pallets/price-feed/src/lib.rs
```diff
@@ -76,6 +76,16 @@ pub mod pallet {
     pub type AssetFeeds<T: Config> =
         StorageMap<_, Blake2_128Concat, T::AssetId, FeedIdFor<T>, OptionQuery>;
 
+    #[pallet::storage]
+    /// (AssetId) -> AssetPricePair
+    ///
+    /// This storage stores the initial price pair for quote assets based on `SelfAssetId`
+    ///
+    /// * insert: adding a new asset with no price pair been set yet
+    /// * remove: an asset has its price pair, remove automatically
```
mattsse, line 85: Should we keep this around, since this is only a single entry?
removing it is probably not worth it

pallets/asset-index/src/lib.rs
```diff
@@ -190,6 +190,10 @@ pub mod pallet {
         ) -> DispatchResultWithPostInfo {
             T::AdminOrigin::ensure_origin(origin.clone())?;
             let caller = ensure_signed(origin)?;
+
+            // Store intial price pair if not exists
+            T::PriceFeed::ensure_price(asset_id, Price::from_inner((value / units).into()))?;
```
mattsse, line 195: would safe math make sense here?

mattsse (approved):
LGTM, minor nits

pallets/price-feed/src/lib.rs
clearloop: Agree with keeping this without removing

I don't have a certain idea about this actually, this confused me a lot since submitting a new feed automatically while adding a new asset is not suitable

pallets/asset-index/src/lib.rs
clearloop: Had noticed but forgot

pallets/price-feed/src/lib.rs
mattsse: yeah, chainlink feeds are kept independently from the assets, and the price pair pallet merely keeps a mapping.
you can also have a look at acala's chainlink integration https://github.com/AcalaNetwork/Acala/pull/1073/files

PR #107, opened 2021-06-10
chore: upgrade to polkadot 0.9.4

mattsse:
## Changes

- update runtime and node to reflect latest changes
- patches substrate/polkadot/cumulus deps
- Use compact Cargo.toml for every pallet

The patches are necessary:
- patching `chainlink-feed` pallet
- ~~patching `jsonrpc*-alpha` versions that always get pulled in and break stuff~~

## Tests

```
cargo t
```

## Issues

- Closes #103

PR #113, opened 2021-06-14
feat: integrate parachain collator pallet with session and cumulus aura

### LINKED ISSUE #109: add pallet session support
## Issue summary

- this will be required in order to configure the aura authorities 

- Ref #107 #103

mattsse:
## Changes

- Adds support for managing collators in a parachain
- Implement Session and Authority for runtime

## Tests

```
cargo test
```

## Issues

Closes #109 Ref #107 

-


node/src/command.rs
```diff
@@ -32,10 +28,9 @@ fn load_spec(
     Ok(match id {
         "pint-local" => Box::new(chain_spec::pint_local_config(para_id)),
         "pint-dev" => Box::new(chain_spec::pint_development_config(para_id)),
-        path => Box::new(
-            chain_spec::ChainSpec::from_json_file(std::path::PathBuf::from(path))
-                .unwrap_or(chain_spec::pint_development_config(para_id)),
-        ),
+        path => Box::new(chain_spec::ChainSpec::from_json_file(
+            std::path::PathBuf::from(path),
+        )?),
```
clearloop, line 33: This will occurs

```
Error: Input("Relay chain argument error: Invalid input: Error opening spec file: No such file or directory (os error 2)")
```

When we pass no `chainspec.json` or `--chain <id>`, personally think running `pint-local` by default is better

clearloop (commented):
Why not adding `pallet-collator-selection` in `Cargo.toml` as a dependency directly?

node/src/command.rs
mattsse: I see, this was defaulted to `.unwrap_or(chain_spec::pint_development_config(para_id)),` which I agree is more convenient.

mattsse:
> Why not adding `pallet-collator-selection` in `Cargo.toml` as a dependency directly?

Oh wow, somehow I missed that this landed on cumulus itself.
Yes, I agree, makes more sense to use this directly.

clearloop (approved):
LGTM now

PR #135, opened 2021-06-22
refactor(asset-index): use user multicurrency accounts when adding assets

mattsse:
## Changes

The `add_asset` extrinsic will be executed following a government proposal to exchange a given number of asset for PINT. This essentially mints new PINT and transfers the asset from the caller to the treasury. 
- This refactoring makes sure that the the caller has enough assets when converting to PINT in their account. 
- Removes the necessity for the `IndexAssetData` since the units are now tracked in the treasury account for the given asset.
- `Holdings` becomes `Assets` and merely serves as an registry for all the known assets.
- It's expected that the caller's account will be funded via an XCM

### TODO
~~The saft registry calls `add_asset` internally. Since SAFTs are minted, this probably should just issue the given amount first before calling `add_asset` to let the caller convert them for PINT?~~

## Tests

```
cargo t --all-features
```


clearloop (approved):
> The saft registry calls add_asset internally. Since SAFTs are minted, this probably should just issue the given amount first before calling add_asset to let the caller convert them for PINT?

What about just allowing adding SAFT assets in `saft_registry` and adding liquid assets in `asset_index`?

We can configure the storage of `asset_index` in `saft_registry`

mattsse:
> > The saft registry calls add_asset internally. Since SAFTs are minted, this probably should just issue the given amount first before calling add_asset to let the caller convert them for PINT?
> 
> What about just allowing adding SAFT assets in `saft_registry` and adding liquid assets in `asset_index`?
> 
> We can configure the storage of `asset_index` in `saft_registry`‘s `Config`

This is great, I think splitting the `add_asset` function into `add_asset` and something like `register_asset` would be appropriate. so the saft index can register the new SAFT and mint them accordingly.

mattsse:
> > > The saft registry calls add_asset internally. Since SAFTs are minted, this probably should just issue the given amount first before calling add_asset to let the caller convert them for PINT?
> > 
> > 
> > What about just allowing adding SAFT assets in `saft_registry` and adding liquid assets in `asset_index`?
> > We can configure the storage of `asset_index` in `saft_registry`‘s `Config`
> 
> This is great, I think splitting the `add_asset` function into `add_asset` and something like `register_asset` would be appropriate. so the saft index can register the new SAFT and mint them accordingly.

I opted for a slightly different solution and extended the `add_asset` functionality to included token transfer and index token minting combined. as well as setting the availability.
Now, for SAFT we need to mint them first before we can add them to the index.

clearloop (approved):

PR #152, opened 2021-06-29
feat: optimise xcm asst management and monitor balances 

### LINKED ISSUE #128: Optimize asset management for bonding and unbonding
## Issue summary

Initially (un)bond xcm calls were intended to be executed for each deposit/withdrawal. However this has several downsides:
* it's inefficient with high Tx volume to execute a single xcm for each trade.
* especially unbonding is limited to 32 concurrent unlocked chunks before funds have to be withdrawn with withdraw_unbonded. Also incoming DOT deposits (bond) and DOT redemptions (unbond) would cancel each other out constantly.

This could be improved so that the PINT chain keeps a certain amount of stash (free balance) on the relay chain to cover final withdrawals. Since we keep track of the free balance and staked amount we can should only be required to unbond if a certain threshold of redemptions/stash is reached.

- Ref #119

mattsse:
## Changes

- introduce the `MinimumRemoteStashBalance` type which is essentially a map (AssetId -> Balance) that determines the minimum amount of balance of a xcm asset should be kept in stash (unbonded) at all times.
- This must be at least the ExistentialDeposit as configured on the asset's native chain (e.g. DOT/Polkadot)
- This will restrict the amount the council can stake
- This will restrict asset withdrawals from PINT (xcm transfers) if the updated stash balance (of the PINT's parachain account) would fall below the configured threshold.

Note: 
* this does not prevent the user from redeeming PINT for assets, but potentially restricts withdrawals from PINT -> parachain/relaychain until the stash exceeds the threshold again.
* it is assumed that the total issuance/total balance of an asset reflects the total balance of the PINT parachain account on the asset's native chain

## Tests

```
cargo t --all-features
```

Follow up tests after #154

## Issues

- Closes #128

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5462,6 +5462,7 @@ dependencies = [
  "frame-support",
  "frame-system",
  "log",
+ "orml-traits",
  "pallet-balances",
  "pallet-staking",
  "parity-scale-codec",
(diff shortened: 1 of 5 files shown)


clearloop (approved):
LGTM

PR #154, opened 2021-06-30
feat: add xcm simulator support

### LINKED ISSUE #70: Add testing support for XCM

mattsse:
## Changes

- Add support for https://github.com/shaunxw/xcm-simulator for testing relay/parachain setups
- introduced `test-utils/xcm-test-support` that exposes a basic relay chain config and helper types. This crate is not included in the workspace, instead it should be included directly so that the runtime can be extended and the `relay_ext` can be overriden.

## Tests

```
cargo t 
```

## Issues

- depends on #152
- Closes #70


pallets/remote-asset-manager/src/tests.rs
```diff
@@ -0,0 +1,2 @@
+// Copyright 2021 ChainSafe Systems
+// SPDX-License-Identifier: LGPL-3.0-only
```
clearloop, line 2: Are there any tests for the XCM test utils that show it works basically?

mattsse: have a look at the tests of the simulator example
https://github.com/shaunxw/xcm-simulator/blob/master/example/src/main.rs#L115
basically it sets up a connection between relay and para so that messages can be exchanged. So you would send a message from para -> relay and then check that this message was executed as expected on relay.

mattsse: @clearloop I'm not too sure about this direct import solutions. this is necessary because in relay.rs we depend the XcmRouter that is generated in the decl_test_network! macro.
I guess it will become more obvious once we add xcm tests to other pallets.

clearloop (approved):
LGTM

pallets/remote-asset-manager/src/tests.rs
clearloop: I'm feeling okay with the 

```
#[path = "..."]
mod xx;
```

solution if it works, we need the test util, but don't have to develop it, soon there will be a better solution

PR #160, opened 2021-07-05
feat(e2e): pend and display errors in e2e tests

### LINKED ISSUE #81: Implement e2e testing
## Implementation details

Implement e2e testing to ensure everything works collectively (e.g. register assets, oracle setup, adding/depositing/withdrawing assets).

### LINKED ISSUE #147: e2e tests for xcm calls
## Issue summary

- [ ] asset-index
- [x] remote-asset-manager

- #81

clearloop:
## Changes

- Add API tests for 
  - [x] setMetadata
  - [x] deposit
  - [x] withdraw
  - [x] complete_withdraw
  - [ ] remove_asset
- Display transaction errors
- Pend errors in queue

## Tests

```
CI
```

## Issues
Ref https://github.com/open-web3-stack/open-runtime-module-library/issues/499
Ref #81  
Ref #147
Cloeses #170 #176 #177 #178 #180

### DIFF
--- a/js/pint-types-bundle/index.ts
+++ b/js/pint-types-bundle/index.ts
@@ -61,6 +61,11 @@ export const definitions = {
                     member: "CommitteeMember",
                     vote: "Vote",
                 },
+                OrmlAccountData: {
+                    free: "Balance",
+                    frozen: "Balance",
+                    reserved: "Balance",
+                },
                 PendingRedemption: {
                     initiated: "BlockNumber",
                     assets: "Vec<AssetWithdrawal>",
--- a/js/pint-types-bundle/index.ts
+++ b/js/pint-types-bundle/index.ts
@@ -13,7 +13,6 @@ export const definitions = {
                 LookupSource: "MultiAddress",
                 LookupSourceFor: "LookupSource",
                 Action: "Call",
-                AssetId: "u32",
                 AccountIdFor: "AccountId",
                 AccountBalance: {
                     available: "Balance",
@@ -25,6 +24,12 @@ export const definitions = {
                         Saft: null,
                     },
                 },
+                AssetId: "u32",
+                AssetMetadata: {
+                    name: "BoundedString",
+                    symbol: "BoundedString",
+                    decimals: "u8",
+                },
                 AssetWithdrawal: {
                     asset: "AssetId",
                     state: "RedemptionState",
@@ -32,6 +37,7 @@ export const definitions = {
                 },
                 Balance: "u128",
                 BalanceFor: "Balance",
+                BoundedString: "BoundedVec<u8, 50>",
                 CommitteeMember: {
                     account_id: "AccountId",
                     member_type: "MemberType",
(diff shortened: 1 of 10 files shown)


js/pint-types-bundle/index.ts
```diff
@@ -25,13 +24,20 @@ export const definitions = {
                         Saft: null,
                     },
                 },
+                AssetId: "u32",
```
mattsse, line 27: can we somehow use the resources/types.json file? so we only have to maintain the types in one spot?

pallets/price-feed/src/lib.rs
```diff
@@ -244,10 +244,16 @@ pub mod pallet {
...
+            let (base_feed_id, quote_feed_id) = match (
+                Self::get_asset_feed_id(&base),
+                Self::get_asset_feed_id(&quote),
+            ) {
+                (None, Some(_)) | (Some(_), None) | (None, None) => {
+                    return <InitialPricePairs<T>>::get(&quote)
+                        .ok_or(Error::<T>::AssetPriceFeedNotFound.into())
+                }
+                (Some(b), Some(q)) => (b, q),
+            };
```
mattsse, line 256: this is rather complex so we should add a comment that explains this.

pallets/price-feed/src/lib.rs
```diff
@@ -244,10 +244,16 @@ pub mod pallet {
...
-                Self::get_asset_feed_id(&quote).ok_or(Error::<T>::AssetPriceFeedNotFound)?;
+            let (base_feed_id, quote_feed_id) = match (
+                Self::get_asset_feed_id(&base),
+                Self::get_asset_feed_id(&quote),
+            ) {
+                (None, Some(_)) | (Some(_), None) | (None, None) => {
+                    return <InitialPricePairs<T>>::get(&quote)
+                        .ok_or(Error::<T>::AssetPriceFeedNotFound.into())
+                }
+                (Some(b), Some(q)) => (b, q),
```
mattsse, line 255: since we have only 2 separate arms here, can we simplify this with `if let (Some(b), Some(q)) = (Self::get_asset_feed_id(&base), Self::get_asset_feed_id(&quote))`?

mattsse (changes requested):
Minor nits,
also, we're constructing the calls using the raw polkadot js api, would it be beneficial in the long term if we generate bindings like in the observability repo?

js/pint-types-bundle/index.ts
clearloop: Or a script that can migrate:
-  `js/pint-types-bundle/index.ts` -> `resouces/types.json`
-  `resources/types.json` -> `js/pint-types-bundle/index.ts`
?

mattsse: the benefit of having this directly available would be that a user can copy types.json right away and use the polkadot UI.
But if we eventually roll out this js folder as a standalone folder, we need to separate these anyway.

So I'm in favor of a `js/pint-types-bundle/index.ts -> resouces/types.json` pipeline

clearloop: > the benefit of having this directly available would be that a user can copy types.json right away and use the polkadot UI.

agree, we can embed the script in `/build.rs`

mattsse (approved):
LGTM

Issue #157, opened 2021-07-02
unify converter implemenentations


mattsse:
## Issue summary

- there are two main `Converter` traits:
    - `sp_runtime::traits::Convert<A, B> -> B`
    -  `xcm_exectutor::traits::Convert<A,B> -> Result<B, A>`

The latter is currently used in the remote asset manager and should be removed once #150 is closed. 

## Other information and links

-

mattsse:
closed #161

PR #161, opened 2021-07-05
refactor(xcm): use xtokens for xcm transfers

### LINKED ISSUE #150: Unify LocalAssetTransactor and add_asset call
## Issue summary

- The `LocalAssetTransactor` is responsible for handling incoming xcm with asset deposits and uses the asset-index as asset/token registry. 
- The assetindex::add_asset call requires funds in the user account before adding assets to the index in exchange for PINT
- Currently both of these features block each other because `LocalAssetTransactor` requires the asset being registered and `add_asset` requires funds being available.
- Solution: extract an `register_asset` call from the current `add_asset` function

- #149

mattsse:
## Changes

- use xcm transfer scheme adopted by xtokens pallet for xcm transfers.
- deprecate `xcm-assets` helper crate, this functionality is now completely provided by the upcoming `XcmTransfer` orml trait
- use `sp_runtime::Convert` instead of `xcm::Convert`
- modify `RemoteAssetManager::transfer` to return the consumed weight

## Tests

```
cargo t --all-features
```

## Issues

- Closes #150 #157 #168

- Blocked by https://github.com/open-web3-stack/open-runtime-module-library/issues/537, currently patched with https://github.com/mattsse/open-runtime-module-library


pallets/remote-asset-manager/src/mock.rs
```diff
@@ -442,7 +458,7 @@ pub mod para {
         type MinimumRemoteStashBalance = MinimumRemoteStashBalance;
         type Assets = Currency;
         type XcmExecutor = XcmExecutor<XcmConfig>;
-        type XcmAssets = xcm_assets::XcmAssetExecutor<XcmAssetConfig>;
+        type XcmAssetTransfer = XTokens;
```
clearloop, line 461: So, if we use `xToken` directly, we don't to convert the errors anymore

clearloop (approved):

pallets/remote-asset-manager/src/mock.rs
mattsse: yeah this is now outsourced to xtokens.

PR #164, opened 2021-07-06
feat: implement remove asset

### LINKED ISSUE #24: Implement `remove_asset` API call

mattsse:
## Changes

- Add `remove_asset` extrinsic and update trait abstraction
- SAFTs are burned, liquids are transfered (from the treasury account)

Revive of #73
## Tests

```
cargo t --all-features
```

## Issues

- Closes #24 #2

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -269,13 +269,14 @@ pub mod pallet {
         ) -> DispatchResultWithPostInfo {
             T::AdminOrigin::ensure_origin(origin.clone())?;
             let caller = ensure_signed(origin)?;
-            let recipient = recipient.unwrap_or(caller);
+            let recipient = recipient.unwrap_or_else(|| caller.clone());
 
             // calculate current PINT equivalent value
             let value = Self::calculate_pint_equivalent(asset_id, units)?;
 
             // transfer the caller's fund into the treasury account
             <Self as AssetRecorder<T::AccountId, T::AssetId, T::Balance>>::remove_asset(
+                caller,
                 asset_id,
                 units,
                 value,
@@ -725,14 +726,14 @@ pub mod pallet {
         }
 
         fn remove_asset(
+            who: T::AccountId,
             asset_id: T::AssetId,
             units: T::Balance,
             nav: T::Balance,
             recipient: Option<T::AccountId>,
         ) -> DispatchResult {
-            let treasury = Self::treasury_account();
             ensure!(
-                T::IndexToken::can_slash(&treasury, nav),
+                T::IndexToken::can_slash(&who, nav),
                 Error::<T>::InsufficientDeposit
             );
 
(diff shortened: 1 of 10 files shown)


clearloop:
Excellent @mattsse ~ I was trying to implement this as well, but it took time for me to reorganize the concepts about SAFT and liquid asset

pallets/asset-index/src/traits.rs
```diff
@@ -28,7 +28,14 @@ pub trait AssetRecorder<AccountId, AssetId, Balance> {
...
 
-    fn remove_asset(id: &AssetId) -> DispatchResult;
+    /// Dispatches transfer to move assets out of the index’s account, if a liquid asset is specified, burns the SAFT otherwise.
+    /// Updates the index by burning the given amount of index token.
+    fn remove_asset(
+        id: AssetId,
+        units: Balance,
+        nav: Balance,
+        recipient: Option<AccountId>,
+    ) -> DispatchResult;
```
clearloop, line 38: `saft_registry` also have function `remove_saft` calling this method which terms to remove the total SAFT, should we separate this into two functions or change the `remove_saft` in `saft-registry` ?

pallets/asset-index/src/lib.rs
```diff
@@ -681,7 +724,33 @@ pub mod pallet {
...
+                    asset_id,
+                    units,
+                )?;
+            } else {
+                // burn SAFT by withdrawing from the index
+                T::Currency::withdraw(asset_id, &treasury, units)?;
+            }
+
+            // burn index token accordingly, no index token changes in the meantime
+            T::IndexToken::slash(&treasury, nav);
```
clearloop, line 752: As a result, treasury slash nav after transferring a liquid asset, should we bind these two operations into one? for example, once we transferring some liquid asset from PINT to other chains, the treasury must slash target nav?

pallets/asset-index/src/traits.rs
mattsse: That'd be definitely an improvement to split this in to functions

mattsse:
@clearloop I split add/remove into corresponding asset type specific functions `add_liquid`/`add_saft` and `remove_liquid`/`remove_saft` so that we now have a better separation of concerns and also simplifies things, like the saft registry doesn't depend on the `Currency` type anymore.

Closes #166

pallets/asset-index/src/lib.rs
mattsse: Not sure what you mean, but we actually don't have to do any transfers of the liquid assets because the whole xcm pipeline takes care of that.

mattsse:
Also @clearloop there is currently a logic bug wrt how xcm transfers are handled, if we set the recipient in `remove_asset` this would result in a withdrawal of the recipient's account. 
this will be addressed in #161 after some changes in the xtokens pallet directly.

clearloop (approved):
LGTM! This is hard for me to separate the issues in the `remove_asset` into several new issues, this solved a big problem of mine, I'll merge and fix the e2e tests in #160

Issue #179, opened 2021-07-07
Decouple AssetIdConvert from AssetIndex


mattsse:
## Issue summary

In order to process incoming XCM deposits the `LocalAssetTransactor` requires the `AssetIdConvert` type that converts `MultiAsset` -> `AssetId`, it does so by decoding the `GeneralKey` of the last Junction of the `MultiLocation` and matches this asset against the registered assets in the index.
However new assets only get included in the `AssetIndex` if there are already deposits available in a sovereign account of the user that calls `AssetIndex::add_asset`.
So, cross chain deposits (`Xcm::Deposit`) depend on `AssetIndex::add_asset` and `AssetIndex::add_asset` requires fulfilled `XcmDeposits`. (This only holds for sibling parachains, not the relaychain).

The `AssetRecorder` trait already exposes a function to register new assets, adding a new gov. extrinsic that enables cross chain deposits will fix this circular dependency.

## Other information and links

-

PR #181, opened 2021-07-07
feat: add asset index genesis config

mattsse:
## Changes

- adds support for registering asset location in genesis

~~Currently blocked by missing serde support for MultiLocation~~ 
Use `ParaId` instead and construct the location during genesis build
## Tests

```
cargo t --all-features
```

## Issues

- Ref #179


clearloop (approved):
LGTM

PR #182, opened 2021-07-07
feat: add register asset extrinsic

### LINKED ISSUE #179: Decouple AssetIdConvert from AssetIndex

mattsse:
## Changes

- Adds another extrinsic to register new assets

We could also think about dropping the `location` parameter of the `add_asset` in favor of this new extrinsic?

## Tests

```
cargo t --all-features
```

## Issues

- Closes #179

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -168,6 +168,9 @@ pub mod pallet {
             Option<AccountIdFor<T>>,
             T::Balance,
         ),
+        /// A new asset was registered in the index
+        /// \[Asset, Availability\]
+        AssetRegistered(T::AssetId, AssetAvailability),
         /// A new deposit of an asset into the index has been performed
         /// \[AssetId, AssetUnits, Account, PINTPayout\]
         Deposited(T::AssetId, T::Balance, AccountIdFor<T>, T::Balance),
@@ -213,6 +216,8 @@ pub mod pallet {
         NAVOverflow,
         /// Thrown when to withdrawals are available to complete
         NoPendingWithdrawals,
+        /// Thrown if the asset that should be added is already registered
+        AssetAlreadyExists,
     }
 
     #[pallet::hooks]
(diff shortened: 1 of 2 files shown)


pallets/asset-index/src/lib.rs
```diff
@@ -281,6 +302,30 @@ pub mod pallet {
...
+            origin: OriginFor<T>,
+            asset_id: T::AssetId,
+            availability: AssetAvailability,
+        ) -> DispatchResult {
+            T::AdminOrigin::ensure_origin(origin.clone())?;
+
+            Assets::<T>::try_mutate(asset_id, |maybe_available| -> DispatchResult {
+                // allow new assets only
+                ensure!(
+                    maybe_available.replace(availability.clone()).is_none(),
```
clearloop, line 319: Will this operation apply to the storage?

clearloop, line 319: If asset exists, we should roll back the storage

clearloop (changes requested):

pallets/asset-index/src/lib.rs
mattsse: `try_mutate` only mutates if `Ok` is returned.

clearloop (approved):

Issue #186, opened 2021-07-07
Make XCMError more detailed


clearloop:
## Issue summary

As testing XCM calls in #160, the provided `XCMError` is helpless for debugging

Should we just extend the Xcm Error to the target pallet's error definition?

Suggest to write a proc-macro for extending `#[pallet:error]` with `XcmError`

## Other information and links

-

PR #189, opened 2021-07-07
feat(runtime): add indices of pallets in runtime

clearloop:
## Changes

This is for easy debugging, in case of the conversion of XCM error is tricky, we can transform the XCM Error into `DispatchError` by hand, for example

```rust
DispatchError::Module {
    module,
    index,
    message,
}
```

## Tests

```
CI
```

## Issues

- #186

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -795,44 +795,43 @@ construct_runtime!(
         NodeBlock = opaque::Block,
         UncheckedExtrinsic = UncheckedExtrinsic,
     {
-        System: frame_system::{Pallet, Call, Config, Storage, Event<T>},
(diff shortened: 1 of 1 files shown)


mattsse (changes requested):
we can structure this a little better depending on the purpose of the pallet? so that if we're flexible when we'll add more in the future

See https://github.com/AcalaNetwork/Acala/blob/master/runtime/acala/src/lib.rs#L1624-L1726

clearloop:
> we can structure this a little better depending on the purpose of the pallet? so that if we're flexible when we'll add more in the future

For example, skip 10 pallets every section?

mattsse (approved):
LGTM, 20 pint pallets should be enough -.-

PR #192, opened 2021-07-08
fix: disallow native asset deposits and withdrawals

### LINKED ISSUE #191: Disallow deposits of native currency
## Issue summary

- the deposit function of the asset-index should fail on PINT (native token) deposits.

-

mattsse:
## Changes

- Ensures the given asset is not the native asset for operations that expect liquid / saft

This would probably be caught as well by the `ensure_liquid_asset` function but this way it's handled explicitly.

## Tests

```
cargo t --all-features
```

## Issues

- Closes #191


clearloop (approved):
LGTM

PR #195, opened 2021-07-08
perf(asset-index): explicit arguments of add_asset and deposit

clearloop:
## Changes

- use `amount` instead of `value` in `add_asset`
- use `units` instead of `amount` in `deposit`

## Tests

```
CI
```

## Issues

-


mattsse (approved):
LGTM, this makes it more consistent and concise.

PR #201, opened 2021-07-13
feat(derive): extend xcm errors with proc-macro for reversed usages

### LINKED ISSUE #186: Make XCMError more detailed

### LINKED ISSUE #200: Extend xcm errors with proc-macro
## Issue summary

Extends xcm errors to custom pallet errors

-

clearloop:
## Changes

`#[xcm_error]` to extend custom pallet errors with xcm errors

## Tests

```
CI
```

## Issues

Closes #200 
Closes #186


mattsse:
fyi: https://docs.rs/synstructure/0.12.5/synstructure/

also can you elaborate on the intended functionality, afaik, pallet errors can't have fields?

clearloop:
> afaik, pallet errors can't have fields?

yep

clearloop:
Hard to use proc-macro to expand `Error<T>` because of the limit of `#[pallet::error]` (`#[pallet::error]` is an attribute of `#[pallet]`, the execution order of attribute macro sucks),

I'd like to 
* hard code the XCMErrors into `Error<T>`
* try custom functional-macro 

How do you like these ideas? @mattsse

mattsse:
> Hard to use proc-macro to expand `Error<T>` because of the limit of `#[pallet::error]` (`#[pallet::error]` is an attribute of `#[pallet]`, the execution order of attribute macro sucks),
> 
> I'd like to
> 
> * hard code the XCMErrors into `Error<T>`
> * try custom functional-macro
> 
> How do you like these ideas? @mattsse, I think it's really important of supporting all Errors in XCM for debugging the unstable XCM calls ( I hate XCM calls LOL)

yeah this is rather unpleasant to work with, so I'd propose to bit this on hold for now. not much we can do at the moment

clearloop (commented):
I don't think we need this macro anymore after #161, keep this only if we need convert XcmErrors in pallets

mattsse (approved):
LGTM,
even though we can't directly put it use yet, my guess would be that this will come in handy if need to refactor transacting xcm calls

PR #204, opened 2021-07-13
feat: add convert to liquid dispatchable call

### LINKED ISSUE #165: Implement `convert_to_liquid` functionality for SAFT
## Issue summary

Eventually a SAFT will become a liquid token.

- the asset availability must be updated to the location of the new parachain
- All SAFT records should be pruned
- who should receive the token distribution? 
- do they need to be transferred first from this new parachain to PINT?

mattsse:
## Changes

- Add `convert_to_liquid` that turns a SAFT into a liquid asset
- Sets the given location of the asset and removes the `SaftRecords`

## Tests

```
cargo t --all-features
```

## Issues

- Closes #165


clearloop (approved):
LGTM

PR #207, opened 2021-07-14
refactor: withdrawal fee model

### LINKED ISSUE #206: Add proper Fee type
## Issue summary

- the trading fee is currently a trait and should be replaced with a simple (numerator , denominator) type.

mattsse:
## Changes

- Adds a fixed fee type for withdrawals, from there the final fee should be a configurable function based on the time the assets remained in the index but this needs some further work on how deposits are tracked first.

## Tests

```
cargo t --all-features
```

## Issues

- Closes #206

### DIFF
--- a/primitives/primitives/src/lib.rs
+++ b/primitives/primitives/src/lib.rs
@@ -50,23 +50,62 @@ pub type FeedId = u64;
 pub type Value = u128;
 
 pub mod fee {
-    use codec::{Encode, Decode};
+    use codec::{Decode, Encode};
 
     /// Represents the fee rate where fee_rate = numerator / denominator
     #[derive(Debug, Encode, Decode, Copy, Clone, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


clearloop (approved):
LGTM

primitives/primitives/src/lib.rs
```diff
@@ -48,3 +48,64 @@ pub type FeedId = u64;
...
 /// Value type for price feeds.
 pub type Value = u128;
+
+pub mod fee {
+    use codec::{Decode, Encode};
+
+    /// Represents the fee rate where fee_rate = numerator / denominator
+    #[derive(Debug, Encode, Decode, Copy, Clone, PartialEq, Eq)]
+    #[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+    pub struct FeeRate {
```
clearloop, line 58: Miss `FeeRate` in types.json

mattsse: good catch!
added this

PR #209, opened 2021-07-14
fix(remote-asset-manager): the usages of runtime-benchmarks feature

### LINKED ISSUE #208: Fix docker-build-release in CI
## Issue summary

Broken after #161 

* Could not compile with `cargo build --release` 

-

clearloop:
## Changes

Could not build with `cargo build --release` because of using `runtime-features` directly in `Cargo.toml`

## Tests

```
CI
```

## Issues

Closes #208

### DIFF
--- a/pallets/committee/src/types.rs
+++ b/pallets/committee/src/types.rs
@@ -10,9 +10,6 @@ use frame_support::{
 };
 use frame_system::RawOrigin;
 
-#[cfg(feature = "std")]
-use serde::{Deserialize, Serialize};
-
 #[derive(PartialEq, Eq, Clone, Encode, Decode, RuntimeDebug)]
 /// This represents an instance of a proposal that can be voted on.
 /// It has been proposed and has an assigned nonce.
@@ -31,7 +28,6 @@ impl<T: Config> Proposal<T> {
 }
 
 #[derive(PartialEq, Eq, Clone, RuntimeDebug, Encode, Decode)]
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 /// Defines what sub-type a member belongs to.
 /// Council members are fixed in number and can vote on proposals
 /// Constituent members are unbounded in number but can only veto council proposals
--- a/pallets/remote-asset-manager/Cargo.toml
+++ b/pallets/remote-asset-manager/Cargo.toml
@@ -54,7 +54,7 @@ pallet-balances = { git = 'https://github.com/paritytech/substrate', branch = 'p
 pallet-staking-reward-curve = { git = 'https://github.com/paritytech/substrate', branch = 'polkadot-v0.9.7' }
 pallet-timestamp = { git = 'https://github.com/paritytech/substrate', branch = 'polkadot-v0.9.7' }
 pallet-session = { git = 'https://github.com/paritytech/substrate', branch = 'polkadot-v0.9.7' }
-pallet-collective = { git = 'https://github.com/paritytech/substrate', branch = 'polkadot-v0.9.7', features = ['runtime-benchmarks'] }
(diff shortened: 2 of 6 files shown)


pallets/remote-asset-manager/Cargo.toml
```diff
@@ -109,6 +109,10 @@ runtime-benchmarks = [
     'frame-benchmarking',
     'frame-support/runtime-benchmarks',
     'frame-system/runtime-benchmarks',
+    'pallet-collective/runtime-benchmarks',
+    # polkadot
+    'xcm-builder/runtime-benchmarks',
+    'pallet-xcm/runtime-benchmarks'
```
mattsse, line 115: this probably won't compile because these are `dev` deps only.

pallets/remote-asset-manager/Cargo.toml
```diff
@@ -65,10 +65,10 @@ cumulus-pallet-xcm = { git = "https://github.com/paritytech/cumulus", branch = "
 parachain-info = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.7" }
 
 # polkadot
-xcm-builder = { git = 'https://github.com/paritytech/polkadot', branch = 'release-v0.9.7', features = ['runtime-benchmarks'] }
+xcm-builder = { git = 'https://github.com/paritytech/polkadot', branch = 'release-v0.9.7' }
 polkadot-runtime-parachains = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.7" }
 polkadot-parachain = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.7" }
-pallet-xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.7", features = ['runtime-benchmarks'] }
+pallet-xcm = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.7" }
```
mattsse, line 71: I've added the features here so that we can. compile it with `--all-features`

pallets/committee/src/types.rs
```diff
@@ -31,7 +28,6 @@ impl<T: Config> Proposal<T> {
 }
 
 #[derive(PartialEq, Eq, Clone, RuntimeDebug, Encode, Decode)]
-#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
 /// Defines what sub-type a member belongs to.
```
mattsse, line 31: we need these derives in order to put membersips into genesis configs

mattsse (changes requested):
I'm not sure when the problems with `cargo test --all-features` started, but essentially the missing `runtime-benchmark` function of the `Ensure` trait is the problem, and we need to enable this for the dev deps as well.

pallets/committee/src/types.rs
clearloop: `GeneisConfig` only requires `AccountId` now, we don't need to write `MemberType` in it anymore

pallets/remote-asset-manager/Cargo.toml
clearloop: Once we use these pallets/dependencies with `runtime-benchmarks` feature, we should add them to `dependencies` with `optional = true`

pallets/remote-asset-manager/Cargo.toml
clearloop: ur right

pallets/remote-asset-manager/Cargo.toml
mattsse: I see, that makes sense.

pallets/committee/src/types.rs
mattsse: ah, you're right, that's even better.

mattsse (approved):

Issue #210, opened 2021-07-16
Conflict MultiLocation required between xtokens and remote-asset-manager


clearloop:
## Issue summary

* Required `X2` in `xtokens`

https://github.com/open-web3-stack/open-runtime-module-library/blob/3bf16d6efc8c35039a062748ff20fa6db6e8faa0/xtokens/src/lib.rs#L323

* Required `X1` in `XcmSender`

https://github.com/paritytech/polkadot/blob/8a6af4412ffc6d327775310c9b4ff527f3345958/runtime/common/src/xcm_sender.rs#L30

## Other information and links

-

mattsse:
Can you extend on that?
the xtokens check just makes sure that there is a chain part and a recipient part (e.g. the account where we want to deposit funds in) and passes this to the `XcmExecutor` which executes the XCM accordingly. The `XcmExecutor` will then use `XcmSender` internally.

this `XcmSender` requires X1 because this actually sends the XCM (sends them to exact location of X1)

clearloop:
Just saw the check in `xtokens` is for `receipient`

PR #213, opened 2021-07-17
feat(remote-asset-manager): use chain part of location while sending xcm calls

### LINKED ISSUE #81: Implement e2e testing

### LINKED ISSUE #147: e2e tests for xcm calls

clearloop:
## Changes

- Introduce `chain_location` in asset-registry

## Tests

```
CI
```

## Issues

- Ref #210 
- Closes #147 
- Closes #81


primitives/primitives/src/traits.rs
```diff
@@ -11,6 +11,9 @@ pub trait MultiAssetRegistry<AssetId> {
     /// asset is native from the point of view of the current system
     fn native_asset_location(asset: &AssetId) -> Option<MultiLocation>;
 
+    /// Determines the chain location of the given asset
```
mattsse, line 14: I feel like this should be a bit more concise wrt `native_asset_location`.
Can we extend on this that this only returns the chain_part where as native_asset_location returns POV

pallets/asset-index/src/lib.rs
```diff
@@ -870,6 +870,17 @@ pub mod pallet {
...
 
+        fn chain_location(asset: &T::AssetId) -> Option<MultiLocation> {
+            let location = Self::native_asset_location(asset)?;
+            match (location.first(), location.at(1)) {
+                (Some(Junction::Parent), Some(Junction::Parachain(id))) => {
+                    Some((Junction::Parent, Junction::Parachain(*id)).into())
+                }
+                (Some(Junction::Parent), _) => Some(Junction::Parent.into()),
+                _ => None,
+            }
```
mattsse, line 881: This the same as `chain_part`? https://github.com/open-web3-stack/open-runtime-module-library/blob/55053241229a8e842dbb717c2bb281daa9fbdf04/traits/src/location.rs#L9 
can we use this directly on `location` instead?

mattsse (approved):
Good catch, this is indeed required if we want to transact directly as we'd require direct channels for this. 

minor nits and suggestions

Issue #159, opened 2021-07-05
configure xcm call weights as constants for different runtimes


mattsse:
## Issue summary

- [ ] add weights for xcm calls used on relay chain (kusama/polkadot), can those runtimes (more importantly their weight impls) be accessed in PINT chain spec?
- [ ] check whether the supported xcm calls (proxy, staking) are independent of the input parameters

## Other information and links

- Ref https://github.com/ChainSafe/PINT/pull/155#discussion_r662633737_

clearloop:
Just realize the `staking_configs` and `proxy_configs` should be dynamic in case of supporting different chains.

Maybe a double map

```rust
pub type Proxies<T: Config> = StorageDoubleMap<
    _,
    Blake2_128Concat,
    AssetId,
    Twox64Concat,
    ParaId,
    ProxyConfig,
    ValueQuery,
>;
```

mattsse:
Closing this since this will be supported by polkadot with upcoming xcm benchmarking pipeline directly

PR #215, opened 2021-07-17
feat(xcm-calls): use polkadot weights as default for xcm-calls

clearloop:
## Changes

- Use polkadot weights as default for xcm-calls

## Tests

```
CI
```

## Issues

- Ref #159


primitives/xcm-calls/src/staking.rs
```diff
@@ -228,10 +229,9 @@ pub struct StakingWeights {
     pub withdraw_unbonded: Weight,
 }
 
-impl StakingWeights {
+impl Default for StakingWeights {
```
mattsse, line 232: hesitant against making this `Default` over `polkadot`, because we'll also need to add support for kusama/westend etc. so keeping dedicated function for each runtime is probably better. 
Also this should be moved to dedicated runtime mod later once we ad a pint runtime for kusama/polkadot

mattsse (changes requested):

primitives/xcm-calls/src/staking.rs
clearloop: What about 

```rust
trait XcmCallWeights {
    fn polkadot() -> Self;
}

impl XcmCallWeights for StakingWeights {
  fn polkadot() -> Self {
      Default::default()
  }
}
```

mattsse: this is reasonable, we can put this later in `runtime/common` and add `fn kusama()`
But I'm in favor of dropping `Default`, rather use it directly in `polkadot`.
naming suggestion: `XcmRuntimeCallWeights`?

mattsse (approved):

mattsse (approved):

Issue #173, opened 2021-07-06
simplify xcm transact


mattsse:
## Issue summary

- Use simpler dispatching solution

## Other information and links

- Blocked by  https://github.com/paritytech/xcm-format

mattsse:
This will require query responses which will be introduced in XCM v2 https://github.com/paritytech/polkadot/pull/3629

PR #222, opened 2021-07-21
feat: enable statemint xcm support

### LINKED ISSUE #58: Token transfer from statemint < > pint chain

### LINKED ISSUE #220: add statemint config and to remote-asset manager
## Issue summary

- [ ] dispatchable for interaction with the assets pallets
- [ ] add weight struct for assets pallet

-

mattsse:
## Changes

- Adds dedicated `StatemintConfig` support. This uses the generic `AssetsCall` and `PalletAssetsEncoder` types to dispatch XCM transact to statemint, once enabled
- Add several maintenance extrinsics for enabling/disabling xcm support, update weights and set statemint config 
- Withdraw functions (PINT -> Statemint)

NOTE: due to the way `Xcm::Transact` currently works, there is no way to get the response of a dispatch call of the target chain, so this currently assumes that the call succeeds. This leads to significant caveats, for example `pallet_assets` calls will fail if the beneficiary account does not exist yet (no `existential deposit`), so in order to receive minted statemint PINT (SPINT?) the account that calls `transfer_to_statemint` must have already an account on statemint. This means essentially: if the call fails on statemint, the funds are locked in the `Statemintcustodian` account on PINT. #173 with upcoming xcm request-response will hopefully make this safe.

## Tests

```
cargo t --all-features
```

## Issues

- Closes #220
- Closes #58

### DIFF
--- a/pallets/remote-asset-manager/src/mock.rs
+++ b/pallets/remote-asset-manager/src/mock.rs
@@ -432,6 +432,7 @@ pub mod para {
     parameter_types! {
         pub LockupPeriod: <Runtime as system::Config>::BlockNumber = 0;
         pub MinimumRedemption: u32 = 0;
+        pub MinimumStatemintTransferAmount: Balance = 1;
         pub WithdrawalPeriod: <Runtime as system::Config>::BlockNumber = 10;
         pub DOTContributionLimit: Balance = 999;
         pub TreasuryPalletId: PalletId = PalletId(*b"12345678");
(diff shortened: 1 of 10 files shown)


pallets/remote-asset-manager/src/mock.rs
```diff
@@ -556,3 +610,224 @@ pub mod para {
         }
     );
 }
+
+pub mod statemint {
```
clearloop, line 614: Could we move mock runtimes like this into one crate?

clearloop (approved):

pallets/remote-asset-manager/src/mock.rs
mattsse: that's not trivial because they use the router that gets generated within the `decl_test_network` macro. that's also the reason why import the relay runtime like this.
we could extract it to a separate mod though.

clearloop: Maybe we can write another macro to generate these?

mattsse: not possible currently, because of circular dependencies: parachain runtimes depend on `Router` that gets generated within the `decl_test_network`.
My guess is that `xcm-simulator` will address this, therefor we can hold off on this.

PR #226, opened 2021-07-23
chore: rm unused adapter types

mattsse:
## Changes

- drop unused adapter types that was replaced with `orml_currency` and `orml_xtokens`

## Issues

-


clearloop (changes requested):
`no_std` dependencies

clearloop (approved):
LGTM

Issue #205, opened 2021-07-13
Optimize asset-index (complete)withdrawal and staking related calls


mattsse:
## Issue summary

- The current implementation makes us of `RemoteAssetManager::unbond` that sends XCM and should be replaced with a function that tracks the unbonded balance

- The redemption state machine is rather verbose and unintuitive can be refactored and potentially simplified

- [ ] simplify redemption state machine, which is rather verbose and unintuitive
- [ ] untangle unstaking related xcm calls from `withdraw` functions

## Other information and links

- Ref #228

PR #228, opened 2021-07-26
refactor: split asset withdraw calculation to separate functions

mattsse:
## Changes

- Split the rather complex `withdraw` dispatchable into separate helper functions and introduce more types. This will come in handy once we add rpc support that provides asset distributions

## Tests

```
cargo t --all-features
```

## Issues

- Ref #205


clearloop (approved):
LGTM

PR #231, opened 2021-07-26
refactor(asset-index): improve complete withdrawal and asset redemption

### LINKED ISSUE #205: Optimize asset-index (complete)withdrawal and staking related calls

### LINKED ISSUE #237: Fix asset NAV calculation with pending withdrawals
## Issue summary

While working on #205 I have discovered the following logic error in asset withdrawals:

### Problem: 
The NAV of an asset is determined by the number of assets currently held by the index treasury. `withdraw` burns the index token directly while assets are transferred after lockup period and `complete_withdraw`. 
This means that withdrawals currently only reduce the index tokens issued (by burning them), while the net asset value calculation continues to use assets that are about to be withdrawn, ultimately inflating the index token NAV. 

### Fix
Use only free balance when calculating NAV of an asset and try to reserve the redeemed amount in the treasury account, removing it from future NAV calculations. Since the redemption amount depends on the ratio of the asset's share in the index, reserving should not fail.

-

mattsse:
## Changes

- Overhauls redemption state machine
- simplify withdrawal by (mostly) decoupling it from xcm unbonding procedures
  - There is still some things left to be done to completely untangle unstaking related xcm calls

## Tests

```
cargo t --all-features
```

## Issues

- Closes #205 
- Closes #237


primitives/primitives/src/traits.rs
```diff
@@ -1,8 +1,8 @@
 // Copyright 2021 ChainSafe Systems
 // SPDX-License-Identifier: LGPL-3.0-only
 
-use frame_support::dispatch::DispatchError;
-use frame_support::sp_runtime::DispatchResult;
+use codec::{Decode, Encode};
+use frame_support::{dispatch::DispatchError, sp_runtime::DispatchResult, RuntimeDebug};
 use xcm::v0::{MultiLocation, Outcome};
```
clearloop, line 6:
```suggestion
use sp_std::boxed::Box;
use xcm::v0::{MultiLocation, Outcome};
```

clearloop (changes requested):

clearloop:
It seems that the code in `remote-asset-manager` has not updated together

mattsse:
> It seems that the code in `remote-asset-manager` has not been updated together

good catch, updated the trait impl, should hopefully pass now

clearloop (approved):
LGTM

PR #235, opened 2021-07-27
feat(asset-index): use PINT instead of IndexToken

### LINKED ISSUE #78: Use `PINT` or `IndexToken` in our code documentation?
## Issue summary

When people talk about `PINT`, people regard `PINT` as the token/currency of this project as our wishes.

Currently, we are using a namespace `IndexToken` in our code (docs) which is standard in development but easy to make the concepts mixed up for developers/contributors —— We already have [too many concepts][0] about tokens in PINT, it is necessary to simplify them.

I'm thinking that we can: 

1. Use one namespace only, replace `IndexToken` with `PINT` in the code.
2. Keep two namespaces, doc the `IndexToken` that it is `PINT` we are commonly talking about.

[0]: https://docs.polkadotindex.com/safts-within-the-pint-index

clearloop:
## Changes

- Rename IndexToken to PINT

## Tests

```
CI
```

## Issues

- Closes #78

### DIFF
--- a/pallets/asset-index/src/benchmarking.rs
+++ b/pallets/asset-index/src/benchmarking.rs
@@ -14,7 +14,7 @@ benchmarks! {
         let asset_id = 42_u32.into();
         let caller: T::AccountId = whitelisted_caller();
         let million = 1_000_000u32.into();
-        T::IndexToken::deposit_creating(&caller, million);
+        T::PINT::deposit_creating(&caller, million);
     }: _(
         RawOrigin::Signed(caller.clone()),
         asset_id,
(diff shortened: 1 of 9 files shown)


mattsse:
Is there a specific reason for this?
I hesitate to include this because PINT is ultimately the name of the polkadot currency, and will be different for Kusama. I agree that naming is partly inconsistent, but probably it makes more sense if we use uniformly `IndexToken`?

clearloop:
> Is there a specific reason for this?
> I hesitate to include this because PINT is ultimately the name of the polkadot currency, and will be different for Kusama. I agree that naming is partly inconsistent, but probably it makes more sense if we use uniformly `IndexToken`?

Reasonable

PR #243, opened 2021-08-04
feat: add balance meter trait

### LINKED ISSUE #233: Add Remote Balance Meter type
## Issue summary

currently there is little support to monitor the actual remote balance. Instead, an estimate is made by tracking asset inflows and outflows. however, additional xcm calls can lead to inconsistencies, by consuming weight on the receiver end, so that PINT parachain needs to pay for XCM execution with assets currently held in the index.

as in its current form, this will be based on a somewhat best estimate. Periodically we could send an additional `Order::QueryHolding` along an unbond that will report back to PINT.

In addition this should keep track for the inflow / outflow of assets.
Operations that impact the remote asset balance:
* deposits: stash + deposit
* withdrawals: stash - withdrawal
* staking rewards: depends on the asset and the staking reward configuration
* unbonding: unbonding period is asset dependent (7 days on kusama 28 days on polkadot, potentially more networks), the proposed unbonding period in PINT is also 28 days, which means that the balance meter must anticipate the reserve balances for 28 days. So that we can cover a potential withdrawal after the PINT redemption period is over. The staking pallet only allows a limited number of unbond calls before we need to withdraw funds. Sending an XCM unbond for each PINT redemption is therefore not feasible, and furthermore, since PINTs are redeemed as a distribution of assets, this might not be economical if the user has to pay the fees (for multiple unbonding calls possibly). Instead we should enforce a certain amount of stash as reserve that will never be staked. This reserve should be `total_balance(DOT) - (total_balance(DOT) - balance(DOT, pint treasury)) + liquid buffered amount`, "All asset inflow that is currently held in user accounts on PINT + additional amount that is the buffer and is liquid". The buffered amount increases as soon as the user deposits asset into the index on pint (`asset_index::deposit`). The buffered amount ensures that there is stash available so that we can cover withdrawals after 28 days. `pallet_staking::bond` essentially needs to be withdrawn from the buffered amount.

the payment for xcm trasnact can now no longer be attributed to a single user, instead it should become part of the redemption fee

- This is somewhat related to #218 in the sense that #218 will introduce some `Trader` (`weight -> asset`) type and `BalanceMeter` keep record of the remote balance.

mattsse:
## Changes

- Adds helper trait `BalanceMeter` that helps keeping track of the remote balances

## Tests

```
cargo t --all-features
```

## Issues

- Closes #233


clearloop (approved):
LGTM

PR #249, opened 2021-08-05
feat: abstract all NAV related calculations over NavProvider trait

### LINKED ISSUE #248: Abstract NAV logic over reusable trait
## Issue summary

- all NAV calculations are embedded in the `asset-index` pallet, it maybe useful to abstract these functions over a trait (`NavProvider`?) so that this can be reused
- the implementation could remain in the `asset-index` since some util function require access to the `Assets` map, or the `MultiAssetRegistry` can be extended to provide this.

- NAV info can be used to determine the necessary reserve of an asset based on it's current value, as the value is reflected in the amounts that are about to be withdrawn: #245

mattsse:
## Changes

- Introduce `NavProvider` that abstracts all NAV calculations and explains all terminology and formula
- Fixes some soundness issues when performing NAV calculations

The `NavProvider` comes with dedicated functions for `SAFT`, `liquid` and `index token`, perhaps it makes sense to further split that?

## Tests

```
cargo t --all-features
```

## Issues

- Closes #248 
- Follow up #250

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -53,7 +53,7 @@ pub mod pallet {
         AssetAvailability, AssetMetadata, AssetRedemption, AssetVolume, AssetWithdrawal,
         AssetsDistribution, AssetsVolume, IndexTokenLock, PendingRedemption, RedemptionState,
     };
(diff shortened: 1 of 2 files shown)


clearloop (approved):
LGTM!

Issue #252, opened 2021-08-06
Simplify asset price feeds


mattsse:
## Issue summary

In order to fetch the necessary price data, calls to the pallet which stores the pricing data will require that an asset pair identifier (eg. ACA/DOT) be provided. 
For the purpose of the PINT Index these will all be using a single denominating asset which will be the one base currency. The asset whose price is being queried will be the quote currency (ie. base/quote).

which leads to the following requirements for the price pallet: 
- [ ] All price feeds must also use the configured base price
- [ ] Price must be for exactly 1 unit

Due to the way the NAV is calculated, NAV will be measured in the configured `base`. Since this will be chainlink feeds, the `base` asset be USD. Right now the `asset-index` always queries a relative price, which is obsolete if NAV is measured in USD. 

## Other information and links

-

PR #257, opened 2021-08-06
refactor: use consistent nav terminology

mattsse:
## Changes

- use consisting terminology and clear some things up
- calculate the nav per token iteratively 
- Use `FixedU128` type for `NAV` since this is essentially the price for 1 unit of index token.
- Use `U256` when aggregating all the asset value. These aren't actually used internally because the NAV calculation is executed iteratively. These are potentially useful to expose via RPC.
- also adds more benchmarks
- move `asset-index` traits to primitives as they are shared with other pallets. 
- Introduce `SaftRegistry` trait for `saft-registry` pallet that gives us all the value secured by all SAFTs. the reason we need this in the `asset-index` is that we need this info to calculate the nav. Now we have a circular dependency that is fixed by moving all the traits to `primitives`. `asset-index` uses `SaftRegistry`, which is implemented on `saft-registry` and `saft-registry` requires `AssetRecorder` which is implemented by `asset-index`. We could move the map that stores all the accumulated SAFTs to `asset-index` but this doesn't fit semantically.

Maybe it makes more sense to add this into the `PriceFeed` pallet so that pallet can also return prices for assets backed by SAFT?

*Note:* this is not actually integrated in the `asset-index` yet because the `PriceFeed` needs some refactoring first #252 . After that all the old nav functions can be replaced.

depends on #258 .

## Tests

```
cargo t --all-features
```

## Issues

-


clearloop (approved):
LGTM

PR #255, opened 2021-08-06
style(rustfmt): add rustfmt.toml && cargo +nightly fmt

mattsse:
## Changes

- Use common config used in substrate/polkadot

## Tests

```

```

## Issues

- Closes #254

PR #258, opened 2021-08-06
feat(saft): use saft identifiers instead of indices

### LINKED ISSUE #250: Keep track of the total SAFT NAV
## Issue summary

- The NAV of a SAFT is determined once a new record is added and remains static for the time being. SAFT's don't have price feeds yet. The NAV can only be changed using `SAFT::registry::report_nav`
- it is therefore difficult to distinguish between SAFT and liquid assets. however, since SAFT are purely static, the total value could be tracked quite easily as additional aggregated value within the `SAFTRegistry::ActiveSaft` map

-

### LINKED ISSUE #256: Keep SAFT record indices consistent
## Issue summary

- single SAFT are identified by their index in the list of `SAFTRecords` in the `ActiveSAFTs` storage.
- SAFTs can be removed entirely; currently, this removes the `SAFTRecord` from the vec, shrinking the vec.
- Therefore an `index` emitted via the `SAFTAdded` event  may no longer correspond to the record.

- Ref #250

mattsse:
## Changes

- introduce unique and constant identifiers for SAFT
- track total NAV of all SAFTs for an asset
- update benchmarks and tests

builds on #255 

## Tests

```
cargo t --all-features
```

## Issues

- Closes #250 
- Closes #256

### DIFF
--- a/primitives/primitives/src/types.rs
+++ b/primitives/primitives/src/types.rs
@@ -2,10 +2,10 @@
 // SPDX-License-Identifier: LGPL-3.0-only
 
 use frame_support::sp_runtime::{
-    app_crypto::sp_core,
-    generic,
-    traits::{BlakeTwo256, IdentifyAccount, Verify},
-    MultiSignature, OpaqueExtrinsic as UncheckedExtrinsic,
+	app_crypto::sp_core,
+	generic,
+	traits::{BlakeTwo256, IdentifyAccount, Verify},
+	MultiSignature, OpaqueExtrinsic as UncheckedExtrinsic,
 };
 
 /// Some way of identifying an account on the chain. We intentionally make it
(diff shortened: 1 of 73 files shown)


clearloop (commented):
Conflicts after #255

mattsse:
> Conflicts after #255

merged main

primitives/primitives/src/types.rs
```diff
@@ -56,4 +56,7 @@ pub type Signature = MultiSignature;
 /// Value type for price feeds.
 pub type Value = u128;
 
+/// Identifier for a SAFT
+pub type SAFTId = u32;
```
clearloop, line 60: New type missed in `types.json`

clearloop (approved):

primitives/primitives/src/types.rs
mattsse: Good catch,
Can we add this to the checklist aswell, I've got a feeling that this will happen a lot more

clearloop: The `e2e` tests check this automatically =.=

PR #259, opened 2021-08-09
feat: simplify price feeds and redemption process

### LINKED ISSUE #237: Fix asset NAV calculation with pending withdrawals

### LINKED ISSUE #252: Simplify asset price feeds

mattsse:
## Changes

- fixed all proportion and redemption math
- extend `NavProvider` with proportion calculation
- add more docs for provider trait

## Tests

```
cargo t --all-features
```

## Issues

- Closes #252 
- Ref #237

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4953,10 +4953,12 @@ dependencies = [
  "parity-scale-codec",
  "polkadot-parachain",
  "primitives",
+ "rand 0.8.4",
  "serde",
  "sp-core",
  "sp-io",
  "sp-runtime",
+ "sp-std 3.0.0",
  "xcm",
  "xcm-executor",
 ]
--- a/pallets/asset-index/Cargo.toml
+++ b/pallets/asset-index/Cargo.toml
@@ -34,14 +34,16 @@ orml-traits = { git = 'https://github.com/open-web3-stack/open-runtime-module-li
 [dev-dependencies]
 serde = { version = "1.0.101" }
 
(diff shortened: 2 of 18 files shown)


clearloop (approved):
LGTM

PR #265, opened 2021-08-17
feat: improve remote staking

### LINKED ISSUE #244: Handle staking related xcm calls in bulk for all pending withdrawals
## Issue summary

Currently, all `AssetWithdrawal` in `complete_withdrawal` are handled separately. this means, if an `unbond` in the `withdraw` call, which has been executed before, has failed it must be executed again. Accumulating the total volume first and then executing a single xcm for the asset will safe a lot of fees.

While this optimization would work on a per user basis, it would lead to stark delays in the withdrawal process, if we'd try to unbond only after the user tried to complete their withdrawal, since the unbonding would then just begin. So instead of handling failed unbonding procedures in `complete_withdrawal`, the failures and their amount should be tracked in the remote asset manager and periodically retried. 

This is a follow up for #205

-

### LINKED ISSUE #263: Simplify the RedemptionState machine
## Issue summary

Before initiated index withdrawals can be completed, a few conditions must be met:
- the redemption period must be over
- the remote balance can cover parachain xcm withdrawals.

Initiating a withdrawals succeeds regardless of the current remote balance. 
Since the withdrawal undergoes a redemption period, the actual assets only need to be available by this point/block.
Therefore we need to track all pending withdrawals and ensure that we can cover the xcm transfer once the period is over.
#244 and #245 describe why and how we would deal with unbonds, leading to an untangling of the whole xcm unbond process from individual withdrawals, so that from pov of asset-index the remote-asset manager merely functions as the tracker of all remote balances. So that the asset-index merely informs the manager about requested withdrawals and then basically waits until the redemption period is over. Meanwhile, the asset manager tracks incoming and outgoing assets xcm volume and reserves the total amount of outstanding redemptions on the remote free balance, either limiting new xcm `staking::bond` or issuing `staking::unbond` and `staking::withdraw` to remove staked assets. This will limit staking related xcm calls and simplify the redemption process.

mattsse:
## Changes

- Keep track of all unlocking chunks
- Use the chain's unbonding duration
- More docs on staking in general

## Tests

```
cargo t --all-features
```

## Issues

- Ref #263
- Ref #244


clearloop (approved):
LGTM

PR #282, opened 2021-08-19
feat: simplify redemption statemachine

### LINKED ISSUE #263: Simplify the RedemptionState machine

mattsse:
## Changes

- This simplifies the state mache for redemption, basically letting the remote asset manager do all the work after #244 

## Tests

```
cargo t --all-features
```

## Issues

- Closes #263

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -31,7 +31,6 @@ pub mod pallet {
 	use frame_support::{
 		dispatch::DispatchResultWithPostInfo,
 		pallet_prelude::*,
-		require_transactional,
 		sp_runtime::{
 			traits::{AccountIdConversion, AtLeast32BitUnsigned, CheckedAdd, CheckedDiv, CheckedSub, Saturating, Zero},
 			ArithmeticError, FixedPointNumber,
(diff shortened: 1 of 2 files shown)


clearloop (approved):
LGTM

PR #284, opened 2021-08-19
feat: handle xcm staking calls on a per block basis

### LINKED ISSUE #244: Handle staking related xcm calls in bulk for all pending withdrawals

mattsse:
## Changes

- This handles xcm staking related calls on a per block basis, rather than on a per deposit/withdrawal basis
- upon block initialization we check whether we need to dispatch a xcm which should succeed because we expect only valid assets to be available for staking. The fees for the weight should be paid for by the LP upon deposit/withdrawal where worst case is 1 deposit /withdrawal is 1 XCM.

This requires some additional work regarding automated rebond/withdraw and still leaves some room for simplifications with the existing xcm extrinsics, which I'll follow up on a new PR. As well as advanced testing.

## Tests

```
cargo t --all-features
```

## Issues

- Closes #244

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -456,6 +456,9 @@ pub mod pallet {
 			// mint index token in caller's account
 			Self::do_mint_index_token(&caller, index_tokens);
 
+			// tell the remote asset manager that assets are available to bond
+			T::RemoteAssetManager::deposit(asset_id, units);
+
 			Self::deposit_event(Event::Deposited(asset_id, units, caller, index_tokens));
 			Ok(())
 		}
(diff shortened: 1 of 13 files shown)


clearloop (approved):
LGTM

PR #288, opened 2021-08-20
feat(asset-index): mint asset into the treasury account in add_asset call

### LINKED ISSUE #276: Mint asset when adding them via governance in add_asset
## Issue summary

As discussed the Governance committee decides the tokens that comprise the index, as well as the allocation of each, which will be set via `add_asset`, therefor `add_asset` should be treated as a minting operation for both PINT and the given asset, rather than a transfer as it's handled in `deposit`.
- [ ] mint PINT into the origin's account
- [ ] mint asset into the treasury account

-

clearloop:
## Changes

- [x] mint PINT into the origin's account
- [x] ~~mint asset into the treasury account~~ ( This has been implemented )

## Tests

```
CI
```

## Issues

- Closes #276

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -290,6 +290,9 @@ pub mod pallet {
 		/// Caller's balance is updated to allocate the correct amount of the
 		/// IndexToken. If the asset does not exist yet, it will get
 		/// created with the given location.
+		///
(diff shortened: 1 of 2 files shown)


mattsse (approved):
lgtm,
can we add some docs to `add_asset`:
*the Governance committee decides the tokens that comprise the index, as well as the allocation of each and their value*

mattsse (approved):

Issue #234, opened 2021-07-26
Tracking issue statemine xcm revamp


mattsse:
## Issue summary

We aim to provide dedicated support for PINT as reserve on statemine/statemint. Currently transferring PINT from PINT parachain to Statemint is experimental and works like this:
1. user wants to transfer x amount of PINT to statemint
2. lock x in custodian account on PINT and mint x on statemint by transacting `pallet_assets::Mint`. 

while transacting call like this is supported with Xcm, it comes with some caveats and limitations:
* transacting calls on behalf of the sending parachain's origin requires direct channels, or potentially a `Xcm::relay` via relay chain.
* existential deposit limitations (requires user already has account on statemint)

This was implemented like this on an experimental basis, but with upcoming statemint changes this will be updated

Eventually statemint will allow parachains to send asset to other parachains and will function as a token registry.

To support statemint as registry for PINT we need to revamp statemint interaction once xcm with other chains is supported. 
Currently only the relay chain governance origin can create assets on Statemint/statemine.

# Tasks 
- [x] support parachain-to-parachain asset transfers via statemint (statemint will use reserve based transfer scheme with standard Xcm withdraw/deposit variants)
- [x] synchronize PINT balances:
  - [x] transfer PINT -> statemint: minting on statemint
- [x] upgrade the `AssetIdConvert` type depending on how statemint marks different assets in `MultiAsset` or `MultiLocation`. (mapping `Statemint::pallet_assets::AssetId` -> `PINT::AssetId`)
- [x] remove the Assets call encoder

## Other information and links

-

PR #291, opened 2021-08-20
feat: update statemint xcm support

mattsse:
## Changes

- No more Xcm Transact when interacting with statemint
- uses native xcm withdraw/deposit scheme when interacting with statemint and drop xcm assets transact support
- dedicated support for transferring tokens to statemint

## Tests
Disabled the statemint test that requires an update to latest 0.9.9

```
cargo t --all-features
```

## Issues

- Ref #234

### DIFF
--- a/js/pint-types-bundle/index.ts
+++ b/js/pint-types-bundle/index.ts
@@ -130,7 +130,6 @@ export const definitions = {
                     withdraw_unbonded: "Weight",
                 },
                 StatemintConfig: {
-                    assets_config: "AssetConfig",
                     parachain_id: "u32",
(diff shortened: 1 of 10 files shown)


clearloop (approved):
LGTM

js/pint-types-bundle/index.ts
```diff
@@ -130,7 +130,6 @@ export const definitions = {
                     withdraw_unbonded: "Weight",
                 },
                 StatemintConfig: {
-                    assets_config: "AssetConfig",
```
clearloop, line 133: 👍

PR #294, opened 2021-08-25
feat(asset-index): timestamp deposits

### LINKED ISSUE #270: Timestamp asset-index deposits
## Issue summary

While the fee model has not been finalized yet, it will depend on how long LP contributions remained in the index. Therefore we need to timestamp (blocknumber) and record the LP’s AccountId is whenever PINT is minted to restrict the redemption of that PINT to the LP and to calculate fees associated with early redemption of the PINT.

- [ ] This can be solved by an additional storage map in `asset-index` (AccountId -> Vec<(Amount PINT, BlockNumber)>). in order to limit these contribution we should opt for a `DepositLimit` which restricts how many deposits can be active, as a precaution, so that we can use a `BoundedVec<(Amount PINT, BlockNumber), DepositLimit>`.  Since we already restrict the amount of allowed deposits, this list should not grow to large.
- [ ] upon redemption we need to consolidate the oldest deposits (first in the Vec) and remove those which are completely withdrawn and update the oldest in the list.
- [ ] Add Fee playholder trait, which should have the form of `redemption_fee(duration: BlockNumber, amount: Balance)` this will come on top of the base fee for the time being.

-

clearloop:
## Changes

- [x] adds storage `Deposits`
- [x] adds config `DepositsLimit`
- [x] check deposits in `deposit`
- [x] remove deposits  in `complete_withdrawal`
- [x] fee placeholder
- [x] tests for above

## Tests

```
CI
```

## Issues

- Closes #270

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -601,12 +601,8 @@ pub mod pallet {
 		/// as soon as the aforementioned conditions are met, regardless of
 		/// whether the other `AssetWithdrawal`s in the same `PendingWithdrawal` set
 		/// can also be closed successfully.
-<<<<<<< HEAD
-		#[pallet::weight(10_000)] // TODO: Set weights
 		#[transactional]
-=======
 		#[pallet::weight(T::WeightInfo::complete_withdraw())] // TODO: Set weights
->>>>>>> main
 		pub fn complete_withdraw(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
 			let caller = ensure_signed(origin)?;
 
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -262,6 +262,8 @@ pub mod pallet {
 		ExpectedLiquid,
 		/// Thrown when trying to remove liquid assets without recipient
 		NoRecipient,
+		/// Thrown when trying to remove deposits with no deposits
+		NoDeposits,
 		/// Invalid metadata given.
 		BadMetadata,
 		/// Thrown if no index could be found for an asset identifier.
@@ -614,6 +616,24 @@ pub mod pallet {
 				if !still_pending.is_empty() {
 					// still have redemptions pending
 					*maybe_pending = Some(still_pending);
+				}
+				Ok(())
+			})?;
+
+			<Deposits<T>>::try_mutate_exists(&caller, |maybe_depositing| -> DispatchResult {
+				let depositing = maybe_depositing.take().ok_or(<Error<T>>::NoDeposits)?;
+
+				// TODO:
+				//
+				// consolidate and clean deposits
+				let still_depositing: Vec<_> = depositing
+					.into_iter()
+					.filter_map(|(index_tokens, block_number)| Some((index_tokens, block_number)))
+					.collect();
+
+				if !still_depositing.is_empty() {
+					// still have redemptions pending
+					*maybe_depositing = Some(still_depositing);
 				}
 				Ok(())
 			})?;
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -733,6 +733,7 @@ pub mod pallet {
 						if *index_tokens > amount {
 							dim = Some((*index_tokens - amount, *block_number));
 							total_fee += T::RedemptionFee::redemption_fee(*block_number, amount);
+							amount = Default::default();
 						}
 						true
 					}
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -725,21 +725,25 @@ pub mod pallet {
 
 				let mut dim: Option<(T::Balance, T::BlockNumber)> = None;
 				deposits.retain(|(index_tokens, block_number)| {
-					if amount >= *index_tokens {
-						amount -= *index_tokens;
-						total_fee += T::RedemptionFee::redemption_fee(*block_number, *index_tokens);
+					if amount.is_zero() {
+						true
+					} else if amount >= *index_tokens {
+						amount = amount.saturating_sub(*index_tokens);
+						total_fee =
+							total_fee.saturating_add(T::RedemptionFee::redemption_fee(*block_number, *index_tokens));
 						false
 					} else {
-						if *index_tokens > amount {
-							dim = Some((*index_tokens - amount, *block_number));
-							total_fee += T::RedemptionFee::redemption_fee(*block_number, amount);
-							amount = Default::default();
-						}
+						// # SAFTY
+						//
+						// index_tokens are always > 0
+						dim = Some((index_tokens.saturating_sub(amount), *block_number));
+						total_fee = total_fee.saturating_add(T::RedemptionFee::redemption_fee(*block_number, amount));
+						amount = T::Balance::zero();
 						true
 					}
 				});
 
-				if !deposits.is_empty() {
+				if deposits.len() > 0 {
 					if let Some(e) = dim {
 						deposits[0] = e;
 					}
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -140,6 +140,12 @@ pub mod pallet {
 	#[pallet::generate_store(pub (super) trait Store)]
 	pub struct Pallet<T>(_);
 
+	/// Testing storage for RedemptionFee
+	#[cfg(test)]
+	#[pallet::storage]
+	#[pallet::getter(fn last_redemption)]
+	pub type LastRedemption<T: Config> = StorageValue<_, (T::BlockNumber, T::Balance), ValueQuery>;
+
 	/// (AssetId) -> AssetAvailability
 	#[pallet::storage]
 	#[pallet::getter(fn assets)]
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -309,6 +309,8 @@ pub mod pallet {
 		/// This gets thrown if the total supply of index tokens is 0 so no NAV can be calculated to
 		/// determine the Asset/Index Token rate.
 		InsufficientIndexTokens,
+		/// Thrown if deposits reach limit
+		TooManyDeposits,
 	}
 
 	#[pallet::hooks]
@@ -500,7 +502,7 @@ pub mod pallet {
 			<Deposits<T>>::try_mutate(&caller, |deposits| -> DispatchResult {
 				deposits
 					.try_push((index_tokens, <frame_system::Pallet<T>>::block_number()))
-					.map_err(|_| Error::<T>::InsufficientDeposit)?;
+					.map_err(|_| Error::<T>::TooManyDeposits)?;
 				Ok(())
 			})?;
 
@@ -614,7 +616,7 @@ pub mod pallet {
 
 			let current_block = frame_system::Pallet::<T>::block_number();
 
-			PendingWithdrawals::<T>::try_mutate_exists(&caller, |maybe_pending| -> DispatchResult {
+			PendingWithdrawals::<T>::try_mutate_exists(&caller, |maybe_pending| -> DispatchResultWithPostInfo {
 				let pending = maybe_pending.take().ok_or(<Error<T>>::NoPendingWithdrawals)?;
 
 				// try to redeem each redemption, but only close it if all assets could be
@@ -639,10 +641,8 @@ pub mod pallet {
 					*maybe_pending = Some(still_pending);
 				}
 
-				Ok(())
-			})?;
-
-			Ok(Some(10_000).into())
+				Ok(().into())
+			})
 		}
 
 		/// Updates the index token locks of the caller.
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -615,12 +615,12 @@ pub mod pallet {
 		/// can also be closed successfully.
 		#[transactional]
 		#[pallet::weight(T::WeightInfo::complete_withdraw())]
-		pub fn complete_withdraw(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
+		pub fn complete_withdraw(origin: OriginFor<T>) -> DispatchResult {
 			let caller = ensure_signed(origin)?;
 
 			let current_block = frame_system::Pallet::<T>::block_number();
 
-			PendingWithdrawals::<T>::try_mutate_exists(&caller, |maybe_pending| -> DispatchResultWithPostInfo {
+			PendingWithdrawals::<T>::try_mutate_exists(&caller, |maybe_pending| -> DispatchResult {
 				let pending = maybe_pending.take().ok_or(<Error<T>>::NoPendingWithdrawals)?;
 
 				// try to redeem each redemption, but only close it if all assets could be
@@ -645,7 +645,7 @@ pub mod pallet {
 					*maybe_pending = Some(still_pending);
 				}
 
-				Ok(().into())
+				Ok(())
 			})
 		}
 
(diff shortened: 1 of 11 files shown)


pallets/asset-index/src/lib.rs
```diff
@@ -96,6 +96,13 @@ pub mod pallet {
 		/// Type used to identify assets
 		type AssetId: Parameter + Member + AtLeast32BitUnsigned + Copy + MaybeSerializeDeserialize;
 
+		/// Restricts how many deposits can be active
+		#[pallet::constant]
+		type DepositLimit: Get<u32>;
```
mattsse, line 101:
```suggestion
		type MaxActiveDeposits: Get<u32>;
```

pallets/asset-index/src/lib.rs
```diff
@@ -133,11 +140,22 @@ pub mod pallet {
...
+
 	/// (AssetId) -> AssetAvailability
 	#[pallet::storage]
 	#[pallet::getter(fn assets)]
 	pub type Assets<T: Config> = StorageMap<_, Blake2_128Concat, T::AssetId, AssetAvailability, OptionQuery>;
 
+	#[pallet::storage]
+	#[pallet::getter(fn deposits)]
+	pub type Deposits<T: Config> =
+		StorageMap<_, Blake2_128Concat, T::AccountId, Vec<(T::Balance, T::BlockNumber)>, ValueQuery>;
```
mattsse, line 157: We should use `BoundedVec` here with the `DepositLimit` instead of a raw vec

pallets/asset-index/src/lib.rs
```diff
@@ -580,6 +616,11 @@ pub mod pallet {
 						if redemption.end_block >= current_block &&
 							Self::do_complete_redemption(&caller, &mut redemption.assets)
 						{
+							// Consolidate and clean deposits
+							if Self::do_consolidate_deposits(&caller, &redemption).is_err() {
+								consolidate_deposits_failed = true;
+							}
+
```
mattsse, line 623: this should be done in the `withdraw` function.

pallets/asset-index/src/lib.rs
```diff
@@ -665,6 +712,71 @@ pub mod pallet {
 			share.checked_div(&nav).ok_or_else(|| ArithmeticError::Overflow.into())
 		}
 
+		/// Consolidate and clean deposits while completing withdraw
+		fn do_consolidate_deposits(
```
mattsse, line 716: We need to this in `withdraw`, basically the workflow is:
1. call `withdraw` with `amount` of pint
2. subtract base fee (we already do this)
3. now compute the redemption fee and subtract from `amount` as well, this is basically this `consolidition`
   in order to do that we need to iterate over deposits and substract the single `Deposits` from `amount` and accumulate the fees.

mattsse (changes requested):
We can do some simplifications.
First: we only need to `consolidate` once in `withdraw`. That's when we calculate the fee, check how we do this already with the base fee.
The amount the user can withdraw is: `amount`  - `base_fee` -  `redemption fee`

So `consolidate` needs to iterate over the entries in `Deposits` as total amount reached `amount` and `redemption fee` is `redemptionFee += T::redemptionFee(Deposits[0].amount, Deposits[0].amount)`

pallets/asset-index/src/tests.rs
```diff
@@ -386,6 +386,73 @@ fn deposit_fails_on_missing_assets() {
...
+		}
+
+		// TODO:
+		//
+		// if use `assert_noop`, here will be an binary mismatched
+		assert_eq!(
+			AssetIndex::deposit(Origin::signed(ASHLEY), ASSET_A_ID, 1),
+			Err(pallet::Error::<Test>::InsufficientDeposit.into())
+		);
+	})
```
clearloop, line 416:
```
---- tests::deposit_fails_on_exceeding_limit stdout ----                                                                                                                                                                                                                        
thread 'tests::deposit_fails_on_exceeding_limit' panicked at 'assertion failed: `(left == right)`                                                                                                                                                                               
  left: `[95, 243, 216, 242, 47, 19, 165, 208, 149, 43, 185, 254, 224, 130, 209, 11, 219, 8, 106, 220, 42, 30, 114, 7, 151, 157, 146, 53, 157, 46, 145, 146]`,                                                                                                                  
 right: `[116, 116, 4, 93, 158, 244, 113, 114, 64, 255, 63, 245, 186, 255, 217, 213, 212, 144, 25, 157, 244, 97, 1, 67, 190, 70, 245, 126, 201, 27, 120, 195]`', pallets/asset-index/src/tests.rs:372:9   
```

If use `assert_noop`, the error above will appear, not sure why this happen

pallets/asset-index/src/lib.rs
```diff
@@ -464,6 +490,14 @@ pub mod pallet {
 			// tell the remote asset manager that assets are available to bond
 			T::RemoteAssetManager::deposit(asset_id, units);
 
+			// insert new deposit
+			<Deposits<T>>::try_mutate(&caller, |deposits| -> DispatchResult {
+				deposits
+					.try_push((index_tokens, <frame_system::Pallet<T>>::block_number()))
+					.map_err(|_| Error::<T>::InsufficientDeposit)?;
+				Ok(())
```
clearloop, line 498: maybe this error throws inside `try_mutate`?

pallets/asset-index/src/lib.rs
```diff
@@ -470,6 +496,14 @@ pub mod pallet {
 			// tell the remote asset manager that assets are available to bond
 			T::RemoteAssetManager::deposit(asset_id, units);
 
+			// insert new deposit
+			<Deposits<T>>::try_mutate(&caller, |deposits| -> DispatchResult {
+				deposits
+					.try_push((index_tokens, <frame_system::Pallet<T>>::block_number()))
+					.map_err(|_| Error::<T>::InsufficientDeposit)?;
```
mattsse, line 503: Can we make a distinct error here like `TooManyDeposits` or something?

pallets/asset-index/src/lib.rs
```diff
@@ -503,7 +537,10 @@ pub mod pallet {
 			)?;
 
 			// amount = fee + redeem
-			let fee = amount.fee(T::BaseWithdrawalFee::get()).ok_or(Error::<T>::AssetUnitsOverflow)?;
+			let fee = amount
+				.fee(T::BaseWithdrawalFee::get())
+				.ok_or(Error::<T>::AssetUnitsOverflow)?
+				.saturating_add(Self::do_consolidate_deposits(&caller, amount)?);
```
mattsse, line 543: this is great!

pallets/asset-index/src/lib.rs
```diff
@@ -600,9 +638,11 @@ pub mod pallet {
 					// still have redemptions pending
 					*maybe_pending = Some(still_pending);
 				}
+
 				Ok(())
 			})?;
-			Ok(().into())
+
+			Ok(Some(10_000).into())
```
mattsse, line 645: What's this additional weight for?

pallets/asset-index/src/lib.rs
```diff
@@ -673,6 +713,51 @@ pub mod pallet {
...
 		}
 
+		/// Consolidate and clean deposits while completing withdraw
+		fn do_consolidate_deposits(caller: &T::AccountId, mut amount: T::Balance) -> Result<T::Balance, DispatchError> {
+			let withdrawal_amount = amount.clone();
+			<Deposits<T>>::try_mutate_exists(&caller, |maybe_depositing| -> Result<T::Balance, DispatchError> {
+				let depositing = maybe_depositing.take().ok_or(<Error<T>>::NoDeposits)?;
+				let mut start: T::BlockNumber = 0_u32.into();
+
+				let still_depositing: Vec<_> = depositing
```
mattsse, line 723: `deposits[0]` is always the oldest and we only take until `amount` is reached. 
can we youse `deposits.retain` here instead?

pallets/asset-index/src/lib.rs
```diff
@@ -673,6 +713,51 @@ pub mod pallet {
...
+							index_tokens = index_tokens.saturating_sub(amount);
+							amount = Default::default();
+							Some((index_tokens, block_number))
+						}
+					})
+					.collect();
+
+				if !still_depositing.is_empty() {
+					// still have redemptions pending
+					*maybe_depositing = Some(still_depositing.try_into().map_err(|_| <Error<T>>::InsufficientDeposit)?);
```
mattsse, line 747: we can use `from_unchecked` here since we know length is ok.

pallets/asset-index/src/lib.rs
```diff
@@ -673,6 +713,51 @@ pub mod pallet {
...
+				}
+
+				if amount != Default::default() {
+					return Err(<Error<T>>::InsufficientDeposit.into());
+				}
+
+				Ok(T::RedemptionFee::redemption_fee(
+					frame_system::Pallet::<T>::block_number().saturating_sub(start),
+					withdrawal_amount,
+				))
```
mattsse, line 757: we need to call this for every entry in `deposits`, so execute this in the loop, because the fee is dependent on how long the deposit spent in the index. And then we need to summarize the total fee.

pallets/asset-index/src/lib.rs
```diff
@@ -673,6 +713,51 @@ pub mod pallet {
...
+				let still_depositing: Vec<_> = depositing
+					.into_iter()
+					.filter_map(|(mut index_tokens, block_number)| {
+						// check depositing duration
+						if block_number < start || start == 0_u32.into() {
+							start = block_number;
+						}
+
+						// consolidate deposits
+						if block_number > frame_system::Pallet::<T>::block_number() {
```
mattsse, line 732: checking the block number here doesn't make sense here
we need to subtract the `index_tokens` from `amount`, 
if `amount` > `index_tokens` we delete the deposit entry and add `totalfee+=redemption_fee(index_tokens, block_number)`
else if `index_tokens` > `amount` we update value and `totalfee+=redemption_fee(index_tokens -amount, block_number)`

mattsse (changes requested):
Getting there, but the consolidation logic needs some changes.

pallets/asset-index/src/lib.rs
```diff
@@ -574,13 +613,14 @@ pub mod pallet {
...
 		/// can also be closed successfully.
+		#[transactional]
 		#[pallet::weight(T::WeightInfo::complete_withdraw())]
 		pub fn complete_withdraw(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
 			let caller = ensure_signed(origin)?;
 
 			let current_block = frame_system::Pallet::<T>::block_number();
 
-			PendingWithdrawals::<T>::try_mutate_exists(&caller, |maybe_pending| -> DispatchResult {
+			PendingWithdrawals::<T>::try_mutate_exists(&caller, |maybe_pending| -> DispatchResultWithPostInfo {
```
mattsse, line 623: Doesn't this work?

```suggestion
			PendingWithdrawals::<T>::try_mutate_exists(&caller, |maybe_pending| -> Result<(), DispatchError> {
```

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
 			share.checked_div(&nav).ok_or_else(|| ArithmeticError::Overflow.into())
 		}
 
+		/// Consolidate and clean deposits while completing withdraw
+		fn do_consolidate_deposits(caller: &T::AccountId, mut amount: T::Balance) -> Result<T::Balance, DispatchError> {
+			<Deposits<T>>::try_mutate_exists(&caller, |maybe_depositing| -> Result<T::Balance, DispatchError> {
+				let mut depositing = maybe_depositing.take().ok_or(<Error<T>>::NoDeposits)?;
+				let mut total_fee: T::Balance = Default::default();
```
mattsse, line 724:
```suggestion
				let mut total_fee: T::Balance = T::Balance::zero();
```

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
...
+
+				let mut dim: Option<(T::Balance, T::BlockNumber)> = None;
+				depositing.retain(|(index_tokens, block_number)| {
+					if amount >= *index_tokens {
+						amount -= *index_tokens;
+						total_fee += T::RedemptionFee::redemption_fee(*block_number, *index_tokens);
```
mattsse, line 730: Use Safe Math, just to be sure

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
 			share.checked_div(&nav).ok_or_else(|| ArithmeticError::Overflow.into())
 		}
 
+		/// Consolidate and clean deposits while completing withdraw
+		fn do_consolidate_deposits(caller: &T::AccountId, mut amount: T::Balance) -> Result<T::Balance, DispatchError> {
+			<Deposits<T>>::try_mutate_exists(&caller, |maybe_depositing| -> Result<T::Balance, DispatchError> {
+				let mut depositing = maybe_depositing.take().ok_or(<Error<T>>::NoDeposits)?;
+				let mut total_fee: T::Balance = Default::default();
+
+				let mut dim: Option<(T::Balance, T::BlockNumber)> = None;
```
mattsse, line 726: This is the remainder, right?
can we name this `rem` or `surplus or something like that?

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
...
+				depositing.retain(|(index_tokens, block_number)| {
+					if amount >= *index_tokens {
```
mattsse, line 728: This will also be true for `amount == index_tokens == 0` which should is guaranteed to never happen, right? Can we at least add a note here or better: `if amount.is_zero() {return true}`

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
...
+						amount -= *index_tokens;
+						total_fee += T::RedemptionFee::redemption_fee(*block_number, *index_tokens);
+						false
+					} else {
+						if *index_tokens > amount {
+							dim = Some((*index_tokens - amount, *block_number));
+							total_fee += T::RedemptionFee::redemption_fee(*block_number, amount);
+							amount = Default::default();
```
mattsse, line 736: We can clean this up a bit,
```
*index_tokens = *index_tokens - amount;
```
Then there is also no need for the `dim` at all, right?

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
...
+						}
+						true
+					}
+				});
+
+				if depositing.len() > 0 {
+					if let Some(e) = dim {
+						depositing[0] = e;
+					}
```
mattsse, line 745: I don't think this is necessary if we modify `index_tokens` within the `retain`, right?

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
...
+						if *index_tokens > amount {
+							dim = Some((*index_tokens - amount, *block_number));
+							total_fee += T::RedemptionFee::redemption_fee(*block_number, amount);
+							amount = Default::default();
+						}
+						true
+					}
+				});
+
+				if depositing.len() > 0 {
```
mattsse, line 742:
```suggestion
				if !depositing.is_empty() {
```

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
...
+					if let Some(e) = dim {
+						depositing[0] = e;
+					}
+					*maybe_depositing = Some(depositing);
+				}
+
+				if amount != Default::default() {
```
mattsse, line 749:
```suggestion
				if !amount.is_zero() {
```

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
 			share.checked_div(&nav).ok_or_else(|| ArithmeticError::Overflow.into())
 		}
 
+		/// Consolidate and clean deposits while completing withdraw
+		fn do_consolidate_deposits(caller: &T::AccountId, mut amount: T::Balance) -> Result<T::Balance, DispatchError> {
+			<Deposits<T>>::try_mutate_exists(&caller, |maybe_depositing| -> Result<T::Balance, DispatchError> {
+				let mut depositing = maybe_depositing.take().ok_or(<Error<T>>::NoDeposits)?;
```
mattsse, line 723: Naming suggestion:
```suggestion
				let mut deposits = maybe_deposits.take().ok_or(<Error<T>>::NoDeposits)?;
```

pallets/asset-index/src/lib.rs
```diff
@@ -677,6 +717,43 @@ pub mod pallet {
 			share.checked_div(&nav).ok_or_else(|| ArithmeticError::Overflow.into())
 		}
 
+		/// Consolidate and clean deposits while completing withdraw
```
mattsse, line 720: Please add some more docs about the intention of this call, can probably use the description in the issue.

pallets/saft-registry/src/mock.rs
```diff
@@ -123,10 +124,16 @@ parameter_types! {
 	pub const BaseWithdrawalFee: primitives::fee::FeeRate = primitives::fee::FeeRate{ numerator: 0, denominator: 1_000,};
 }
 
+pub struct RedemptionFee;
+
+impl primitives::traits::RedemptionFee<BlockNumber, Balance> for RedemptionFee {}
```
mattsse, line 129: Can we instead `impl RedemptionFee for ()`?

mattsse (changes requested):
This looks great,
Some nits about naming and some suggestions for optimizations

pallets/asset-index/src/lib.rs
clearloop: `retain` provides a reference which we can not modify it

mattsse: I see, makes sense, then we only need to store `let mut rem: Option<T::Balance>` and update this like you did already

mattsse (approved):
LGTM.
I added some comments for clarity and make the redemption fee actually take the duration

PR #300, opened 2021-08-30
feat(runtime): use committee origin for pallet asset-index

### LINKED ISSUE #268: Check origins for all pallets
### pallet-xcm

+ [ ] SendXcmOrigin -  `EnsureXcmOrigin<Origin, LocalOriginToLocation>`

### cumulus_pallet_dmp_queue

+ [ ] ExecuteOverweightOrigin - `frame_system::EnsureRoot<AccountId>`

### pallet_collator_selection

+ [ ] UpdateOrigin - `EnsureApprovedByCommittee`

### pallet_local_treasury

+ [ ] AdminOrigin - `frame_system::EnsureRoot<AccountId>`

### pallet_saft_registry

+ [ ] AdminOrigin - `frame_system::EnsureSigned<AccountId>`

### pallet_committee

+ [ ] ProposalSubmissionOrigin - `EnsureSigned<AccountId>`
+ [ ] ProposalExecutionOrigin - `EnsureMember<Self>`
+ [ ] ApprovedByCommitteeOrigin - `EnsureApprovedByCommittee`

### pallet_price_feed

+ [ ] AdminOrigin - `frame_system::EnsureRoot<AccountId>`

### pallet_asset_index

+ [ ] AdminOrigin - `frame_system::EnsureSigned<AccountId>`

### pallet_remote_asset_manager

+ [ ] AdminOrigin - `frame_system::EnsureSigned<AccountId>`

- #104 
- #140

### LINKED ISSUE #302: Supports creating feed in price-feed in feature runtime-benchmarks
## Issue summary

The `deposit` call of asset-index requires exists feed, which we can not add in its benchmarks directly, needs to support this in `runtime-benchmarks` feature

-

clearloop:
## Changes

-

## Tests

```
CI
```

## Issues

- Closes #302 
- #268

### DIFF
--- a/pallets/price-feed/src/lib.rs
+++ b/pallets/price-feed/src/lib.rs
@@ -273,24 +273,33 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			use frame_benchmarking::vec;
 
+			pallet_chainlink_feed::Pallet::<T>::set_feed_creator(
+				<frame_system::Origin<T>>::Signed(pallet_chainlink_feed::Pallet::<T>::pallet_admin()).into(),
+				caller.clone(),
+			)?;
+
 			pallet_chainlink_feed::Pallet::<T>::create_feed(
 				<frame_system::Origin<T>>::Signed(caller.clone()).into(),
-				600u32.into(),
+				100u32.into(),
 				Zero::zero(),
 				(1u8.into(), 100u8.into()),
 				1u8.into(),
-				5u8,
+				8u8,
 				vec![1; T::StringLimit::get() as usize],
 				Zero::zero(),
-				vec![(caller.clone(), caller)],
+				vec![(caller.clone(), caller.clone())],
 				None,
 				None,
 			)?;
 
-			let _ = AssetFeeds::<T>::mutate(&asset_id, |maybe_feed_id| {
-				maybe_feed_id.replace(<pallet_chainlink_feed::FeedCounter<T>>::get())
-			});
-
+			let feed_id = <pallet_chainlink_feed::FeedCounter<T>>::get() - 1.into();
+			AssetFeeds::<T>::insert(&asset_id, feed_id);
+			pallet_chainlink_feed::Pallet::<T>::submit(
+				<frame_system::Origin<T>>::Signed(caller.clone()).into(),
+				feed_id,
+				1_u32.into(),
+				42.into(),
+			)?;
 			Ok(().into())
 		}
 	}
--- a/pallets/price-feed/src/lib.rs
+++ b/pallets/price-feed/src/lib.rs
@@ -41,15 +41,15 @@ mod types;
 pub mod pallet {
 	#[cfg(feature = "runtime-benchmarks")]
 	pub use crate::traits::PriceFeedBenchmarks;
-	pub use crate::{traits::PriceFeed, types::TimestampedValue};
+	#[cfg(feature = "runtime-benchmarks")]
+	use frame_benchmarking::Zero;
 	#[cfg(feature = "std")]
 	use frame_support::traits::GenesisBuild;
+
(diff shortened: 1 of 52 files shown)


clearloop:
This PR simply add an interface for creating feed in pallet price-feed

The benchmarks of asset-index is still broken  with:

```
Error:                                                                                                                                 
* Pallet: AssetIndex                                                                                                                   
* Benchmark: deposit                                                                                                                   
* Lowest_range_values: []                                                                                                              
* Highest_range_values: []                                                                                                             
* Steps: []                                                                                                                            
* Repeat: 1                                                                                                                            
* Verify: true                                                                                                                         
* Error message: An overflow would occur 
```

Which is from the pallet-chainlink-feed, tried `submit` round in the `chainlink-feed` node-template directly, not work as well

pallets/price-feed/src/lib.rs
```diff
@@ -260,6 +265,45 @@ pub mod pallet {
...
+			pallet_chainlink_feed::Pallet::<T>::submit(
+				<frame_system::Origin<T>>::Signed(caller.clone()).into(),
+				feed_id,
+				1_u32.into(),
+				42.into(),
+			)?;
+			Ok(().into())
+		}
+	}
+
```
mattsse, line 306: This is essentially a mock?
I'm rather against adding this. Can you extend on why this is necessary, we should fix this upstream if this is a problem of the chainlink pallet.

mattsse (changes requested):
I don't really understand the price feed benchmark extensions and I don't think we should add this like that because the `PriceFeedBenchmarks` is very invasive. But maybe I didn't get the problem.

mattsse:
Also there is already the `Feedbuilder` we use for price feed tests https://github.com/ChainSafe/PINT/blob/bb27364ebe9cc7021860630aa7502e80bc6b7999/pallets/price-feed/src/mock.rs
can't we use that instead?

clearloop:
> Also there is already the `Feedbuilder` we use for price feed tests https://github.com/ChainSafe/PINT/blob/bb27364ebe9cc7021860630aa7502e80bc6b7999/pallets/price-feed/src/mock.rs
> can't we use that instead?

The dependencies benchmarks depend on are not as same as the tests, they are using the real runtimes, and whatever we have `MockPriceFeed` or `Feedbuilder`, the benchmarks don't care about these, they only work for tests ( including the tests in benchmarks ), but not work for benchmarks, 

So if we want to generate weights for extrinsics like `deposit` in pallet asset-index, we need to `create_feed` and `submit` a round in the pallet asset-index, it used to be workable because we had a cache for the price of a new asset before

pallets/price-feed/src/lib.rs
clearloop: Yep, I'm thinking of fixing it in the chainlink repo and open another issue for this

mattsse:
gotcha,
this is probably solvable because the `PriceFeed` pallet is also `chainlink_feed_pallet::Pallet` so you should be able to access that, and we need to enable this in the runtime section as well?
this could work?

clearloop:
> gotcha,
> this is probably solvable because the `PriceFeed` pallet is also `chainlink_feed_pallet::Pallet` so you should be able to access that, and we need to enable this in the runtime section as well?
> this could work?

Yep, and this `PriceFeedBenchmarks` is for this creating feed requirement, tried to re-generate weights for all pallets, both `asset-index` and `saft-registry` has the overflow problem mentioned above, not certain about why it happens currently now

mattsse:
Okay seems we need to track this down.
Can you please split the committee origin and create a separate PR for this?

clearloop:
> Okay seems we need to track this down.
> Can you please split the committee origin and create a separate PR for this?

Yep, in #305 now =.=

mattsse (approved):
lgtm

PR #309, opened 2021-08-31
feat: add Acala compatible workspace

mattsse:
## Changes

This is intended to be the submodule branch for the Mandala integration
- Remove all unnecessary stuff 

@clearloop should we keep the CI stuff probably yes, right?

## Tests

```

```

## Issues

-


clearloop:
I think we don't have to clean our repo only for the branch of the Acala integration, if the target branch has too much different from our main branch, the updates will be a problem, what about a specific runtime inside the main branch?

polkadot has a lot of branches because it's really big

mattsse:
> I think we don't have to clean our repo only for the branch of the Acala integration, if the target branch has too much different from our main branch, the updates will be a problem, what about a specific runtime inside the main branch?
> 
> polkadot has a lot of branches because it's really big

We only need the pallets and primitives for acala integration, we can only merge those folders, which should be easy with git fetch && git checkout HEAD pallets or something.
What we need to modify is at least the Cargo.toml because this nested workspace will throw an error when used as submodule so we need to get rid of this.

clearloop:
> We only need the pallets and primitives for acala integration, we can only merge those folders, which should be easy with git fetch && git checkout HEAD pallets or something.
> What we need to modify is at least the Cargo.toml because this nested workspace will throw an error when used as submodule so we need to get rid of this.
> 
> So you're suggesting only removing th `Cargo.toml` file?
> I had a look at the other ecosystem modules and they are pretty much cleaned up. so I did the same.

Clean it up is also okay, the only thing I'm considering is about the problem of updating the branch, what about the acala guys writing scripts to modify the workspace section in our `Cargo.toml` =.=

mattsse:
> > We only need the pallets and primitives for acala integration, we can only merge those folders, which should be easy with git fetch && git checkout HEAD pallets or something.
> > What we need to modify is at least the Cargo.toml because this nested workspace will throw an error when used as submodule so we need to get rid of this.
> > So you're suggesting only removing the `Cargo.toml` file?
> > I had a look at the other ecosystem modules and they are pretty much cleaned up. so I did the same.
> 
> Clean it up is also okay, the only thing I'm considering is about the problem of updating the branch, what about the acala guys writing scripts to modify the workspace section in our `Cargo.toml` =.=

There won't be a workspace Cargo.toml, because cargo does not support [nested workspaces](https://doc.rust-lang.org/cargo/reference/manifest.html#the-workspace-section) therefore we only keep the pallets' Cargo.toml file.
But perhaps we could exclude the whole pint subdirectory instead, let me try that first.

mattsse:
you were right, `excluding` the pint workspace was way easier than modifying the whole workspace
we should propably fix #310 on main first before including this.

clearloop:
@mattsse  can you merge main and check if clippy is happy?

clearloop (approved):

mattsse:
closing this since we do not need to make any changes anymore. instead we can branch off main directly.

PR #316, opened 2021-09-01
feat: add MaybeAssetIdConvert trait for runtime benchmarks

### LINKED ISSUE #310: Drop TryFrom for AssetId
## Issue summary

- The TryFrom Restriction is not compatible with Acala afterall so we should remove that.
- The whole problem is that we need a way to create `AssetId` in benchmarks without certain restrictions. We probably should do the same as with the Price feed benchmarks

-

mattsse:
## Changes

- Use a feature gated trait to create AssetIds in runtime benchmarks

## Tests

```
cargo t --all-features
```

## Issues

- Closes #310

### DIFF
--- a/pallets/asset-index/Cargo.toml
+++ b/pallets/asset-index/Cargo.toml
@@ -67,6 +67,7 @@ runtime-benchmarks = [
     'frame-support/runtime-benchmarks',
     'pallet-price-feed/runtime-benchmarks',
 #    'pallet-chainlink-feed/runtime-benchmarks',
+    'primitives/runtime-benchmarks',
 ]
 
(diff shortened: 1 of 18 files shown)


clearloop (approved):
lgtm, could it use for the Acala integration?

mattsse:
> lgtm, could this be used for the Acala integration?

I thought so, but then I realized only if we implement this directly in the `CurrencyId` primitives pallet which is not ideal, because then pint is included everywhere, I guess we'd have to add another trait type which is also not ideal -.-.

mattsse:
This trait will be implemented on the Runtime automatically

clearloop (approved):

PR #319, opened 2021-09-02
Use committee origin for saft-registry and asset-index

### LINKED ISSUE #318: Reset origins for pallet asset-index and pallet saft-registry
## Issue summary

Most of the calls inside these two pallets requires `caller` as parameter, the current implementation of `GovernanceOrigin` actually Either<(), CommitteeOrigin> which could not return `AccountId` when executing extrinsic with `Root` origin.

mb we can just use `CommitteeOrigin` for these two pallets, or trying to construct some new origins like `Either<Sudo, CommitteeOrigin>`?

-

clearloop:
## Changes

- Use CommitteeOrigin instead of GovernanceOrigin

## Tests

```
CI
```

## Issues

- Closes #318

### DIFF
--- a/pallets/committee/src/types.rs
+++ b/pallets/committee/src/types.rs
@@ -189,7 +189,20 @@ impl<O: Into<Result<Origin<T>, O>> + From<Origin<T>> + Clone, T: Config> EnsureO
 
 	#[cfg(feature = "runtime-benchmarks")]
 	fn successful_origin() -> O {
-		O::from(CommitteeOrigin::ApprovedByCommittee(Default::default(), Default::default()))
+		use frame_benchmarking::vec;
+		O::from(CommitteeOrigin::ApprovedByCommittee(
+			Default::default(),
+			VoteAggregate {
+				votes: vec![
+					MemberVote {
+						member: CommitteeMember { account_id: Default::default(), member_type: MemberType::Council },
+						vote: Vote::Aye
+					};
+					T::MinCouncilVotes::get() + 1
+				],
+				end: <frame_system::Pallet<T>>::block_number() + 1_u32.into(),
+			},
+		))
 	}
 }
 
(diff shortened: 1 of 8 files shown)


pallets/asset-index/src/mock.rs
```diff
@@ -169,7 +169,7 @@ impl primitives::traits::RedemptionFee<BlockNumber, Balance> for RedemptionFee {
 }
 
 impl pallet_asset_index::Config for Test {
-	type AdminOrigin = frame_system::EnsureSignedBy<AdminAccountId, AccountId>;
```
clearloop, line 172: We don't need to restrict `EnsureSignedBy` in tests since this is only work for tests

pallets/committee/src/types.rs
```diff
@@ -189,7 +189,20 @@ impl<O: Into<Result<Origin<T>, O>> + From<Origin<T>> + Clone, T: Config> EnsureO
...
+			VoteAggregate {
+				votes: vec![
+					MemberVote {
+						member: CommitteeMember { account_id: Default::default(), member_type: MemberType::Council },
+						vote: Vote::Aye
+					};
+					T::MinCouncilVotes::get() + 1
+				],
+				end: <frame_system::Pallet<T>>::block_number() + 1_u32.into(),
+			},
```
mattsse, line 204: Can you please add a quick test that checks that is indeed successful?

mattsse (approved):
Awesome!
I'd suggest a test for successfull_origin.
This means we should now also add additional `force_` extrinsics that are signed by root?

pallets/committee/src/types.rs
clearloop: emmm

This seems only can be tested with benchmarks =.=, we can mock an OuterOrigin to test this but only in tests

This modification works on my local machine to run benchmarks, for testing this, I think the better way is to write a CI try executing benchmarks in our repo

mattsse: you're right, since it works in the benchmarks already, should be fine

PR #320, opened 2021-09-02
feat: add default remote asset manager impl

mattsse:
## Changes

- Add a default implementation for remote asset manager that does nothing

## Tests

```

```

## Issues

-


clearloop (approved):
lgtm

PR #334, opened 2021-09-06
feat: make --dev work standalone

### LINKED ISSUE #333: Make dev runtime independent of relay chain
## Issue summary

- The current node setup requires relay chain setup even for dev runtime
- The dev or pint-dev runtime should work as standalone chain with aura and optional instant sealing.

-

mattsse:
## Changes

- This ensures that we can run `./target/release/pint --dev --instant-sealing `

## Tests

```

```

## Issues

- Closes #333


mattsse:
Somehow blocks are not finalized, am I missing something @clearloop ?

clearloop:
Should produce/finalize by txs

mattsse:
This instant-seal for dev setup ups.
Tx are included regardless.

clearloop:
We need to deposit some balances to accounts like `Alice` so that we can run txs, we could not test with the current dev config because no account have balances

mattsse:
yes, we need to increase the default balance (substrate default is 1 << 60), should we add this?
I'm working on a PR that describes how to run with custom chain spec json, that should be enough to tinker with

clearloop:
> yes, we need to increase the default balance (substrate default is 1 << 60), should we add this?
> I'm working on a PR that describes how to run with custom chain spec json, that should be enough to tinker with

Yep, at least deposit some balances to `Alice` which is set as root account? And then we at least can start with the sudo call `set_balance()`

clearloop:
will this break `polkadot-launch` and the e2e tests since it is using `--chain=pint-dev`

mattsse:
could be, 
now that we have a chain spec included, should I rather revert #8558a11 ?

clearloop:
> could be,
> now that we have a chain spec included, should I rather revert #8558a11 ?

I'm testing on my local machine now, seems no balances in the polkadot.js UI, have you tried to connect the node with UI?

clearloop:
> could be,
> now that we have a chain spec included, should I rather revert #8558a11 ?

no need, e2e tests can follow this by set_balances first

node/src/service.rs
```diff
@@ -69,15 +71,15 @@ pub trait IdentifyVariant {
...
 	}
 
 	fn is_polkadot(&self) -> bool {
-		self.id().starts_with("polkadot")
+		self.id().contains("polkadot")
 	}
 
 	fn is_dev(&self) -> bool {
-		self.id().starts_with("dev")
+		self.id().contains("dev")
```
clearloop, line 82: this will match 'pint-dev' as 'dev' as well and blocking 'polkadot-launch' and e2e tests connecting to the relay chain, see the finalization failed in CI

clearloop (approved):

node/src/service.rs
mattsse: I see, so we should check if this is equal to `pint_dev` instead?

node/src/command.rs
```diff
@@ -296,10 +310,20 @@ pub fn run() -> Result<()> {
...
 			let runner = cli.create_runner(&cli.run.normalize())?;
+			let chain_spec = &runner.config().chain_spec;
+			let is_pint_dev = chain_spec.is_dev();
+
+			set_default_ss58_version(chain_spec);
 
 			runner.run_node_until_exit(|config| async move {
 				let para_id = chain_spec::Extensions::try_get(&*config.chain_spec).map(|e| e.para_id);
 
+				if is_pint_dev {
```
clearloop, line 321: In this checking

mattsse: should we `self.id() == "pint_dev"` instead?

node/src/command.rs
```diff
@@ -296,10 +310,20 @@ pub fn run() -> Result<()> {
 		}
 		None => {
 			let runner = cli.create_runner(&cli.run.normalize())?;
+			let chain_spec = &runner.config().chain_spec;
+			let is_pint_dev = chain_spec.is_dev();
```
clearloop, line 314: What about 
```suggestion
                      let is_pint_dev = cli.base.base.shared_params.dev; 
```

mattsse: this will probably not work for `chain pint-dev`

mattsse: ah I think I can follow now.
yes lets make this change

mattsse:
now we enable instant sealing if `--instant-sealing` or the `--dev` was provided. `--pint-dev` will work as before.

mattsse:
Okay, judging by the e2e tests, it looks like some accounts now just need some funding?

clearloop:
> Okay, judging by the e2e tests, it looks like some accounts now just need some funding?

I'll fix it

PR #336, opened 2021-09-06
feat: add index token deposit range

### LINKED ISSUE #271: Enforce Deposit Limits
## Issue summary

- Add a check to ensure sufficient contributions. There is already `DOTContributionLimit` which was intended to restrict the maximum amount of that DOT that can exist in the index, instead this should enforce both a minimum and maximum amount. This should be checked after determining the pint equivalent in `deposit`

-

mattsse:
## Changes

- Add a configurable range for valid deposits

## Tests

```

```

## Issues

- Closes #271


clearloop (approved):
LGTM

PR #390, opened 2021-09-15
feat(asset-index): add check for decimals of asset metadata

### LINKED ISSUE #363: TOB-PINT-3 - Data validation
Missing decimal validation in the set_metadata function.
Description
The function set_metadata in figure 3.1 doesn't validate the decimals variable lower and upper bound. This allows the governance committee to set the decimals of a given asset to 0 or to a very high value, which are undesired.

The set_metadata function (PINT/pallets/asset-index/src/lib.rs#L472-L493

clearloop:
## Changes

- introduce `MaxDecimals`

## Tests

```
CI
```

## Issues

- Closes #363

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -104,7 +104,7 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxActiveDeposits: Get<u32>;
 
-		/// Restricts how many deposits can be active
+		/// Restricts the max limit of decimals in metadata
 		#[pallet::constant]
 		type MaxDecimals: Get<u8>;
 
@@ -328,9 +328,9 @@ pub mod pallet {
 		InsufficientIndexTokens,
 		/// Thrown if deposits reach limit
 		TooManyDeposits,
-		/// Thrown when the given DepositRange is invalid
-		InvalidDecimals,
 		/// Thrown when the given decimals is zero or too high
+		InvalidDecimals,
+		/// Thrown when the given DepositRange is invalid
 		InvalidDepositRange,
 		/// The deposited amount is below the minimum value required.
 		DepositAmountBelowMinimum,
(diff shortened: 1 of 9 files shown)


runtime/common/src/constants.rs
```diff
@@ -118,6 +118,7 @@ parameter_types! {
 	pub const TreasuryPalletId: PalletId = PalletId(*b"Treasury");
 	pub const LockupPeriod: BlockNumber = 10;
 	pub const MaxCandidates: u32 = 200;
+	pub const MaxDecimals: u8 = 12;
```
mattsse, line 121: we should bump this to 18

pallets/asset-index/src/lib.rs
```diff
@@ -104,6 +104,10 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxActiveDeposits: Get<u32>;
 
+		/// Restricts how many deposits can be active
+		#[pallet::constant]
+		type MaxDecimals: Get<u8>;
```
mattsse, line 109: This comment is not correct.

pallets/asset-index/src/lib.rs
```diff
@@ -325,6 +329,8 @@ pub mod pallet {
 		/// Thrown if deposits reach limit
 		TooManyDeposits,
 		/// Thrown when the given DepositRange is invalid
+		InvalidDecimals,
+		/// Thrown when the given decimals is zero or too high
```
mattsse, line 333: comments switched

mattsse (changes requested):
I don't think this is a critical feature since we don't actually use the metadata.
But it's nice to have it.

clearloop:
> I don't think this is a critical feature since we don't actually use the metadata.

mb can leave it to the client who builds Apps around PINT

mattsse (approved):
lgtm

PR #400, opened 2021-09-16
feat(weight): update dispatch class for important calls

### LINKED ISSUE #375: Elevate priority of important operations
Make extrinsincs related to governance either `DispatchClass::Mandatory` if possible or `Operational`

clearloop:
## Changes

- [x] updates dispatch class for pallet asset-index
- [x] updates dispatch class for pallet saft-registry
- [x] update dispatch calss for pallet committee

## Tests

```

```

## Issues

- Closes #375


mattsse (changes requested):
Since these calls are executed via governance, we should instead make the `vote` and `close` calls operational.
`add_asset` remove_asset etc can't be called directly, so we don't need the dispatch class here

mattsse (approved):
lgtm

PR #406, opened 2021-09-21
feat(committee): introduce storage VotingTimeouts

### LINKED ISSUE #373: Add timeout mechanism for Eligibility to vote
Currently new constituents are allowed to vote in the same voting period they're added

clearloop:
## Changes

- introduce storage `VotingTimeouts`
- insert new members in to `VotingTimeouts`
- check and clean timeouts on voting
- tests for `MemberNotInVotingPeriod`

## Tests

```
CI
```

## Issues

- Closes #373

### DIFF
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -449,9 +449,9 @@ pub mod pallet {
 				let member = maybe_member.take().ok_or(Error::<T>::NotMember)?;
 
 				// Check if have enough council members
-				if member.0 == MemberType::Constituent
-					|| Members::<T>::iter_values().filter(|(m, _)| *m == MemberType::Council).count()
-						> T::MinCouncilVotes::get()
+				if member.0 == MemberType::Constituent ||
+					Members::<T>::iter_values().filter(|(m, _)| *m == MemberType::Council).count() >
+						T::MinCouncilVotes::get()
 				{
 					Ok(member.0)
 				} else {
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -449,9 +449,9 @@ pub mod pallet {
 				let member = maybe_member.take().ok_or(Error::<T>::NotMember)?;
 
 				// Check if have enough council members
-				if member.0 == MemberType::Constituent ||
-					Members::<T>::iter_values().filter(|(m, _)| *m == MemberType::Council).count() >
-						T::MinCouncilVotes::get()
+				if member.0 == MemberType::Constituent
+					|| Members::<T>::iter_values().filter(|(m, _)| *m == MemberType::Council).count()
+						> T::MinCouncilVotes::get()
 				{
 					Ok(member.0)
 				} else {
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -374,14 +374,15 @@ pub mod pallet {
 		pub fn vote(origin: OriginFor<T>, proposal_hash: HashFor<T>, vote: VoteKind) -> DispatchResult {
 			let voter = Self::ensure_member(origin)?;
 
-			if VotingEligibility::<T>::get(&voter.account_id)
-				.filter(|block_number| {
-					*block_number != T::BlockNumber::zero() && *block_number > frame_system::Pallet::<T>::block_number()
+			VotingEligibility::<T>::get(&voter.account_id)
+				.and_then(|block_number| {
+					if frame_system::Pallet::<T>::block_number() >= block_number {
+						Some(block_number)
+					} else {
+						None
+					}
 				})
-				.is_some()
-			{
-				return Err(Error::<T>::NotEligibileToVoteYet.into());
-			}
+				.ok_or(Error::<T>::NotEligibileToVoteYet)?;
 
 			Votes::<T>::try_mutate(&proposal_hash, |maybe_votes| -> DispatchResult {
 				let votes = maybe_votes.as_mut().ok_or(Error::<T>::NoProposalWithHash)?;
@@ -475,9 +476,9 @@ pub mod pallet {
 				let ty = maybe_member.take().ok_or(Error::<T>::NotMember)?;
 
 				// Check if have enough council members
-				if ty == MemberType::Constituent ||
-					Members::<T>::iter_values().filter(|m| *m == MemberType::Council).count() >
-						T::MinCouncilVotes::get()
+				if ty == MemberType::Constituent
+					|| Members::<T>::iter_values().filter(|m| *m == MemberType::Council).count()
+						> T::MinCouncilVotes::get()
 				{
 					VotingEligibility::<T>::take(&member);
 					Ok(ty)
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -125,7 +125,7 @@ pub mod pallet {
 	pub type Votes<T: Config> =
 		StorageMap<_, Blake2_128Concat, HashFor<T>, VoteAggregate<AccountIdFor<T>, BlockNumberFor<T>>, OptionQuery>;
 
-	/// Storte a mapping (hash) -> BlockNumber for new members' voting timeout
+	/// Storte a mapping (hash) -> BlockNumber for new members' voting timeouts
 	#[pallet::storage]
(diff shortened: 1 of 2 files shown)


mattsse (changes requested):
I pictured this a bit differently.
Can we add a maker, like `eligible_at : BlockNumber` to the Member type instead?
this way we can restrict new members very easily from voting in the current period by timestamping their eligibility.

clearloop:
> this way we can restrict new members very easily from voting in the current period by timestamping their eligibility.

This needs to refactor the `Members` storage and a lot of logic related to it and takes more persistent storage for this one period check

mattsse:
> > this way we can restrict new members very easily from voting in the current period by timestamping their eligibility.
> 
> This will take more persistent storage for this one period check

That's a good point. I missed that.
you're absolutely right, since we also use this in `EnsureCommittee` effectively, it would be more performant to use a separate StorageMap for that since this is only required for `propose`.

pallets/committee/src/lib.rs
```diff
@@ -353,24 +361,33 @@ pub mod pallet {
...
-				} else {
-					Err(Error::<T>::NoProposalWithHash)
+			VotingTimeouts::<T>::try_mutate(&voter.account_id, |maybe_timeout| -> DispatchResult {
+				if maybe_timeout
+					.take()
+					.filter(|block_number| {
+						frame_system::Pallet::<T>::block_number().saturating_sub(*block_number) < T::VotingPeriod::get()
+					})
+					.is_some()
+				{
```
mattsse, line 372: this means members included at genesis are missing in the `VotingTimeouts` set, right?
I wonder if we should instead make this conistent for all members in `VotingTimeouts`, so that we track the first block at which they are eligible to vote, which is `0` for members in genesis

mattsse (commented):
This lgtm, only suggestion to use `VotingTimeouts` for every Member and perhaps rename this to `VotingEligibility`?

pallets/committee/src/lib.rs
clearloop: This can make all members equal at the `VotingEligibility` part in logic, but seems have no benefits for solving the voting attack problem since this check if for new members

pallets/committee/src/lib.rs
```diff
@@ -185,6 +194,8 @@ pub mod pallet {
 		/// Attempted to cast a vote outside the accepted voting period for a
 		/// proposal
 		NotInVotingPeriod,
+		/// Attempted to cast a vote without voting eligibility
+		NoVotingEligibility,
```
mattsse, line 198:
```suggestion
		NotEligibileToVoteYet,
```

pallets/committee/src/lib.rs
```diff
@@ -141,12 +147,15 @@ pub mod pallet {
 	#[pallet::genesis_build]
 	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
 		fn build(&self) {
+			let voting_period = T::VotingPeriod::get();
 			for member in &self.council_members {
 				Members::<T>::insert(member, MemberType::Council);
+				VotingEligibility::<T>::insert(member, voting_period);
```
mattsse, line 153: This should be zero so they're instantly able to vote.
And should be documented: Genesis members can vote instantly.
```suggestion
				VotingEligibility::<T>::insert(member, 0);
```

pallets/committee/src/lib.rs
```diff
@@ -141,12 +147,15 @@ pub mod pallet {
...
 			}
 
 			for member in &self.constituent_members {
 				Members::<T>::insert(member, MemberType::Constituent);
+				VotingEligibility::<T>::insert(member, voting_period);
```
mattsse, line 158:
```suggestion
				VotingEligibility::<T>::insert(member, 0);
```

pallets/committee/src/lib.rs
```diff
@@ -351,26 +362,35 @@ pub mod pallet {
 		/// This can only be called by members of the committee.
```
mattsse, line 362:
```suggestion
		/// This can only be called by members of the committee that are eligible to vote.
		/// New members are eligible to vote after 1 voting period has passed from the block they were added to the members set.
```

pallets/committee/src/lib.rs
```diff
@@ -125,6 +126,11 @@ pub mod pallet {
 	pub type Votes<T: Config> =
 		StorageMap<_, Blake2_128Concat, HashFor<T>, VoteAggregate<AccountIdFor<T>, BlockNumberFor<T>>, OptionQuery>;
 
+	/// Storte a mapping (hash) -> BlockNumber for new members' voting timeouts
```
mattsse, line 129:
```suggestion
	/// Stores the block height at which point a member is eligible to cast their vote
	/// For new members, this will be the block they were added as members plus the duration of one voting period.
	/// This means new members are eligible to cast their vote in the next voting period.
```

pallets/committee/src/lib.rs
```diff
@@ -351,26 +362,35 @@ pub mod pallet {
...
-					ensure!(!votes.has_voted(&voter.account_id), Error::<T>::DuplicateVote);
-					votes.cast_vote(MemberVote::new(voter.clone(), vote.clone())); // mutates votes in place
-					Self::deposit_event(Event::VoteCast(voter, proposal_hash, vote));
-					Ok(())
-				} else {
-					Err(Error::<T>::NoProposalWithHash)
+			VotingEligibility::<T>::try_mutate(&voter.account_id, |maybe_eligibility| -> DispatchResult {
+				if maybe_eligibility
+					.take()
+					.filter(|block_number| frame_system::Pallet::<T>::block_number() < *block_number)
```
mattsse, line 373: should we check for `0` here?

mattsse (changes requested):
some nits and suggestions

Also, we should kill the Eligibility after a member is removed

pallets/committee/src/lib.rs
mattsse: Yes, this is just for consistency's sake.

pallets/committee/src/lib.rs
clearloop: here is a `take` before `filter`, `if frame_system::Pallet::<T>::block_number() >= block_number {}`, the storage will be removed automatically

mattsse: the `VotingEligibility::<T>::try_mutate` is unnecessary anyway, isn't it?
we should `get` it instead and then do the check.

clearloop: ur right, used to try using `try_mutate` to avoid unexpected storage modification, since there is a `#[transactional]` tag on this function already, we can take and check directly

mattsse: We don't want to remove the eligibility timestamp. we should just check it instead. so no modifications of the eligibility map.

pallets/committee/src/lib.rs
```diff
@@ -375,7 +375,9 @@ pub mod pallet {
 			let voter = Self::ensure_member(origin)?;
 
 			if VotingEligibility::<T>::get(&voter.account_id)
-				.filter(|block_number| frame_system::Pallet::<T>::block_number() < *block_number)
+				.filter(|block_number| {
+					*block_number != T::BlockNumber::zero() && *block_number > frame_system::Pallet::<T>::block_number()
+				})
```
mattsse, line 380: Flip the if condition instead:

`>= *block {} else {error}`

pallets/committee/src/lib.rs
```diff
@@ -348,33 +363,45 @@ pub mod pallet {
...
-			})?;
+			VotingEligibility::<T>::get(&voter.account_id)
+				.and_then(|block_number| {
+					if frame_system::Pallet::<T>::block_number() >= block_number {
+						Some(block_number)
+					} else {
+						None
+					}
+				})
+				.ok_or(Error::<T>::NotEligibileToVoteYet)?;
```
mattsse, line 385: can we make this:

`.filter(|block_number|frame_system::Pallet::<T>::block_number() >= *block_number)`

instead

mattsse (approved):
last nit

pallets/committee/src/lib.rs
clearloop: forgot that recording the eligibilities now

PR #410, opened 2021-09-22
feat(committee): refactor Proposal and add timeout

### LINKED ISSUE #393: Add timeout for proposals
## Issue summary

- proposals that are accepted must be executed/closed within the next? voting period or they become invalid. 

-

### LINKED ISSUE #408: Remove proposals in ActiveProposals after closing/executing
## Issue summary

for the `ActiveProposals` storage in pallet committee, there are only insert logic but no remove logics

- should remove proposals in ActiveProposals after executing/closing

Or

- add status field to the `Proposal` type
- remove `ExecutedProposals`
- remove `ActiveProposals`

clearloop:
## Changes

- [x] introduce `ProposalStatus`
- [x] remove storage `ExecutedProposals`
- [x] add timeout
- [x] update types in js kit

## Tests

```
CI
```

## Issues

- Closes #408
- Closes #393

### DIFF
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -328,7 +328,7 @@ pub mod pallet {
 
 			// Create a new proposal with a unique nonce
 			let nonce = Self::take_and_increment_nonce()?;
-			let proposal = Proposal::<T>(nonce.clone(), *action);
+			let proposal = Proposal::<T>::new(nonce.clone(), *action, ProposalStatus::Active);
 
 			let proposal_hash = proposal.hash();
 
@@ -394,7 +394,7 @@ pub mod pallet {
 
 			// Execute the proposal
 			let proposal = Self::get_proposal(&proposal_hash).ok_or(Error::<T>::NoProposalWithHash)?;
-			let result = proposal.1.dispatch(Origin::<T>::ApprovedByCommittee(closer, votes).into());
+			let result = proposal.action.dispatch(Origin::<T>::ApprovedByCommittee(closer, votes).into());
 
 			// register that this proposal has been executed
 			ExecutedProposals::<T>::insert(proposal_hash, ());
@@ -444,9 +444,9 @@ pub mod pallet {
 				let ty = maybe_member.take().ok_or(Error::<T>::NotMember)?;
 
 				// Check if have enough council members
-				if ty == MemberType::Constituent ||
-					Members::<T>::iter_values().filter(|m| *m == MemberType::Council).count() >
-						T::MinCouncilVotes::get()
+				if ty == MemberType::Constituent
+					|| Members::<T>::iter_values().filter(|m| *m == MemberType::Council).count()
+						> T::MinCouncilVotes::get()
 				{
 					Ok(ty)
 				} else {
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -244,7 +244,7 @@ pub mod pallet {
 			Proposals::translate_values(|mut proposal: Proposal<T>| {
 				Self::get_votes_for(&proposal.hash()).and_then(
 					|votes: VoteAggregate<T::AccountId, T::BlockNumber>| -> Option<Proposal<T>> {
-						if n.saturating_sub(votes.end) >= voting_period && proposal.status == ProposalStatus::Active {
+						if n.saturating_sub(votes.end) > voting_period && proposal.status == ProposalStatus::Active {
 							proposal.status = ProposalStatus::Closed;
 						}
 
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -195,8 +195,8 @@ pub mod pallet {
 		ProposalNotAcceptedConstituentVeto,
 		/// Tried to close a proposal but proposal was denied by council
 		ProposalNotAcceptedCouncilDeny,
-		/// Attempted to execute a proposal that has already been closed
-		ProposalAlreadyClosed,
+		/// Attempted to execute a proposal that is timeout
+		ProposalTimeout,
 		/// Attempted to execute a proposal that has already been executed
 		ProposalAlreadyExecuted,
 		/// Reach the minimal number of the limit of council members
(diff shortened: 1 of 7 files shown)


pallets/committee/src/lib.rs
```diff
@@ -240,6 +238,21 @@ pub mod pallet {
...
+						if n.saturating_sub(votes.end) > voting_period && proposal.status == ProposalStatus::Active {
+							proposal.status = ProposalStatus::Closed;
+						}
+
+						Some(proposal)
+					},
+				)
+			})
+		}
 	}
```
mattsse, line 256: This is a hook to check timed proposals, right?
it probably would be more efficient to only run this once per voting period?
also what does `translate_values` do exactly?
Wouldn't it be more efficien to keep the `ExecutedProposals` map and loop over the `Proposals` map here? this way we could get rid of the ProposalState

pallets/committee/src/lib.rs
```diff
@@ -199,6 +195,8 @@ pub mod pallet {
 		ProposalNotAcceptedConstituentVeto,
 		/// Tried to close a proposal but proposal was denied by council
 		ProposalNotAcceptedCouncilDeny,
+		/// Attempted to execute a proposal that has already been closed
+		ProposalAlreadyClosed,
 		/// Attempted to execute a proposal that has already been executed
 		ProposalAlreadyExecuted,
```
mattsse, line 201: What's the difference between closed and executed?
This is rather confusing since the extrinsic is also named `closed`

mattsse (changes requested):
From what I'm understanding `Closed` would be "timed out" right?
The hook could be optimized

pallets/committee/src/lib.rs
```diff
@@ -380,33 +393,41 @@ pub mod pallet {
...
+				let current_block = frame_system::Pallet::<T>::block_number();
 
-			// Ensure voting has accepted proposal
-			votes.is_accepted(T::MinCouncilVotes::get()).map_err(Into::<Error<T>>::into)?;
+				// Ensure voting period is over
+				ensure!(current_block > votes.end, Error::<T>::VotingPeriodNotElapsed);
 
-			// Execute the proposal
-			let proposal = Self::get_proposal(&proposal_hash).ok_or(Error::<T>::NoProposalWithHash)?;
-			let result = proposal.1.dispatch(Origin::<T>::ApprovedByCommittee(closer, votes).into());
```
mattsse, line 397: shouldn't this be the original issuer of the proposal?

pallets/committee/src/lib.rs
clearloop: > it probably would be more efficient to only run this once per voting period?

There are different voting period for different proposals, for example, as we set `VotingPeriod` to `10`, Alice submits a proposal at block 1, the voting period is `1~11`, Bob submits a proposal at block 3, the voting period will be `3~13`, we could not run this once per voting period because we don't we when people submit proposals

> also what does translate_values do exactly?

https://substrate.dev/rustdocs/latest/frame_support/storage/types/struct.StorageMap.html#method.translate_values

> Wouldn't it be more efficient to keep the ExecutedProposals map and loop over the Proposals map here? this way we could get rid of the ProposalState

If keep `ExecutedProposals`, there will need another storage map `ClosedProposals` or `Timeout?Proposals` because the expired proposals are not executed, otherwise we can modify `ExecutedProposals` to `ClosedProposals`.

For why we need to make ExecutedProposals and `ClosedProposals` different, plz take a look at https://polkadot.polkassembly.io/

pallets/committee/src/lib.rs
clearloop: https://github.com/ChainSafe/PINT/pull/410#discussion_r714061234

pallets/committee/src/lib.rs
clearloop: https://github.com/ChainSafe/PINT/issues/373#issuecomment-919839363 should be another pr since this requires to fix the tests and benchmarks together

pallets/committee/src/lib.rs
clearloop: Or execute this in the `close` function, if expired close but not execute

mattsse: > > it probably would be more efficient to only run this once per voting period?
> 
> From the exist implementation, there are different voting periods for different proposals, for example, as we set `VotingPeriod` to `10`, Alice submits a proposal at block 1, the voting period is `1~11`, Bob submits a proposal at block 3, the voting period will be `3~13`, we could not run this once per voting period because we don't know when people submit proposals

I see, makes sense.

> > also what does translate_values do exactly?
> 
> https://substrate.dev/rustdocs/latest/frame_support/storage/types/struct.StorageMap.html#method.translate_values
> 
> > Wouldn't it be more efficient to keep the ExecutedProposals map and loop over the Proposals map here? this way we could get rid of the ProposalState
> 
> If keep `ExecutedProposals`, there will need another storage map `ClosedProposals` or `Timeout?Proposals` because the expired proposals are not executed, otherwise we can modify `ExecutedProposals` into `ClosedProposals` directly.
> 
> For why we need to make `ExecutedProposals` and `ClosedProposals` different, plz take a look at https://polkadot.polkassembly.io/


Do we really need to close them automatically? Since we know that they're timed out we could instead fail on `close`, this would simplify things quite a bit. The the state of the proposal makes perfect sense.

pallets/committee/src/lib.rs
mattsse: This terminology is a bit unfortunate because the `fn close` tries to _execute_ the proposal. 
how about TimedOut?

pallets/committee/src/lib.rs
clearloop: Closing them automatically cost a lot and really dirty indeed, but imaging a UI of PINT, people query status of proposals, they'll get `Active` for proposals which are timeout when the committee haven't close them by hand.

> Since we know that they're timed out we could instead fail on close, this would simplify things quite a bit

I'm considering what about we close timeout proposals without throwing errors in the close function?

```
if proposal is timeout
  mark_timetout();
else
  execute() && mark_executed()
```

mattsse (approved):
lgtm.
But we should definitely follow up with some additional tests for timeout out etc.

clearloop:
> lgtm.
> But we should definitely follow up with some additional tests for timeout out etc.

mb some tests mocking the voting attack?   

- submit a proposal
- add new members
- new members try to vote this proposal
  - has no eligibility
- when new members can vote to the proposal
  - proposal timeout

PR #419, opened 2021-09-27
feat(committee): testing new members cannot vote

### LINKED ISSUE #418: More tests for the timeout of proposals
lgtm.
But we should definitely follow up with some additional tests for timeout out etc.

_Originally posted by @mattsse in https://github.com/ChainSafe/PINT/pull/410#pullrequestreview-762781876_

clearloop:
## Changes

-
-
-

## Tests

```
CI
```

## Issues

- Closes #418

### DIFF
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -462,10 +462,8 @@ pub mod pallet {
 				}
 			})?;
 
-			VotingEligibility::<T>::insert(
-				&constituent,
-				frame_system::Pallet::<T>::block_number() + T::VotingPeriod::get(),
-			);
+			let block_numer = frame_system::Pallet::<T>::block_number();
+			VotingEligibility::<T>::insert(&constituent, block_numer + Self::get_next_voting_period_end(&block_numer)?);
 
 			Self::deposit_event(Event::NewConstituent(constituent));
 			Ok(())
--- a/pallets/committee/src/tests.rs
+++ b/pallets/committee/src/tests.rs
@@ -341,14 +341,15 @@ fn cannot_close_if_council_rejects() {
 fn cannot_close_if_constituents_veto() {
 	new_test_ext(0..4).execute_with(|| {
 		add_constituents(4..8);
+		run_to_block(START_OF_V1);
 
 		let proposal = submit_proposal(123);
+		run_to_block(START_OF_V1 + START_OF_S1 - 1);
 
-		run_to_block(START_OF_S1);
 		vote_with_each(0..4, proposal.hash(), VoteKind::Aye);
 		vote_with_each(4..8, proposal.hash(), VoteKind::Nay);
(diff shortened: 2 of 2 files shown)


pallets/committee/src/lib.rs
```diff
@@ -462,10 +462,8 @@ pub mod pallet {
 				}
 			})?;
 
-			VotingEligibility::<T>::insert(
-				&constituent,
-				frame_system::Pallet::<T>::block_number() + T::VotingPeriod::get(),
-			);
+			let block_numer = frame_system::Pallet::<T>::block_number();
+			VotingEligibility::<T>::insert(&constituent, block_numer + Self::get_next_voting_period_end(&block_numer)?);
```
clearloop, line 466: Dispatch the eligibility with `get_next_voting_period_end` since `Proposals`' end are generated by this function

BTW, not sure what is `ProposalSubmission` for, since different proposals have different voting period, this config terms to be no use

pallets/committee/src/tests.rs
```diff
@@ -386,6 +387,31 @@ fn cannot_execute_proposal_twice() {
...
+		run_to_block(START_OF_V1 - 1);
+		vote_with_each(0..4, proposal.hash(), VoteKind::Aye);
+		assert_noop!(
+			Committee::vote(Origin::signed(CONSTITUENT), proposal_hash, VoteKind::Nay),
+			pallet::Error::<Test>::NotEligibileToVoteYet
+		);
+
+		run_to_block(START_OF_V1);
+		assert_noop!(
+			Committee::vote(Origin::signed(CONSTITUENT), proposal_hash, VoteKind::Nay),
```
clearloop, line 406: `vote` call doesn't check proposal status now, so we can only test the VotingPeriod

mb check proposals' status is better but `Proposal` and `VoteAggregate` just using different storages now

pallets/committee/src/lib.rs
mattsse: submitting a proposal and submitting a vote are two different periods.

mattsse (approved):
lgtm

pallets/committee/src/lib.rs
clearloop: If we keep this `SubmissionPeriod` we should add a check into `propose` call to ensure that people are submitting a proposal in SubmissionPeriod ? Currently there are no checks in `propose` call with `SubmissionPeriod`

pallets/committee/src/tests.rs
clearloop: > > vote call doesn't check proposal status now,
> 
> we do check it is in voting period?

Yep we only check if it is in voting period but no checks for if it is closed or timeout, for example, people try to vote and accept `NotInVotingPeriod` but they will never now if the proposal has been closed or timeout in this response

Issue #331, opened 2021-09-03
Review checklist


mattsse:
## Issue summary

- [x] access control
- [x] all dispatchables and complex code paths documented and matches function behaviour
- [x] Safe Match

## Other information and links

-

PR #420, opened 2021-09-27
fix(asset-index): use transfer instead of deposit for liquid assets

### LINKED ISSUE #411: Require liquid assets were sent to via XCM
## Issue summary

- Currently `add_asset` mints PINT + liquid assets a like, this assumes they were contributed. Instead, it should be enforced that these are reserve-backed transferred to PINT first.

-

mattsse:
## Changes

- Fixes a soundness issue mentioned in #331, minting of liquid assets is removed and requires that these assets are deposited via XCM first regardless if it's gov approved.

## Tests

Failing tests are updated to allocate some funds for `add_asset`
Did not check the JS test suite however.

```
cargo t
```

## Issues

- Closes #411


pallets/asset-index/src/lib.rs
```diff
@@ -1034,8 +1035,8 @@ pub mod pallet {
 			}
 			// native asset can't be added
 			Self::ensure_not_native_asset(&asset_id)?;
-			// mint asset into the treasury account
-			T::Currency::deposit(asset_id, &Self::treasury_account(), units)?;
+			// transfer the asset from the caller to treasury account
+			T::Currency::transfer(asset_id, caller, &Self::treasury_account(), units)?;
```
clearloop, line 1039: the caller account may not have enough balance to make transfer

clearloop (changes requested):

pallets/asset-index/src/lib.rs
mattsse: yeah but we catch the error and abort in case that happens?

clearloop: > The given amount of assets must already exist in the caller's account

After modifying this to `transfer`, we don't have any extrinsic left can mint assets in our pallets, for approving an asset working in PINT, we need to mint the assets with orml pallet by hand, or we'll mint them from the XCM part?

mattsse: this is intended as discussed here https://github.com/ChainSafe/PINT/pull/401#discussion_r712791918 
Minting liquid assets without receiving them via XCM is a soundness issue

clearloop:
plz fix the benchmarks

clearloop:
Still broken in the benchmarks, I can join to debug this later

mattsse:
> Still broken in the benchmarks, I can join to debug this later

ah I know what's up, fixing this rn

clearloop (approved):

PR #424, opened 2021-09-28
feat(committee): introduce PendingVotingPeriod

### LINKED ISSUE #422: Add update voting period extrinsic
The new call should only accept values in range of [7, 28], also this should only kick in after the current voting period has ended, so we probably need a storage slot to store the pending voting period until then.

needs a follow up to update the voting period if it's [7, 28] days

_Originally posted by @mattsse in https://github.com/ChainSafe/PINT/pull/421#pullrequestreview-764100449_

clearloop:
## Changes

- [x] adds `PendingVotingPeriod`
- [x] adds tests
- [x] updates benchmarks

## Tests

```
CI
```

## Issues

- Closes #422


clearloop:
Has problem building in docker, not sure if it is related to #425

pallets/committee/src/lib.rs
```diff
@@ -450,10 +454,19 @@ pub mod pallet {
...
+			})?;
 
-				Ok(().into())
-			})
+			// Set new `VotingPeriod` if no active proposals left and `PendingVotingPeriod` is some
+			if Proposals::<T>::iter_values().filter(|p| p.status == ProposalStatus::Active).count().is_zero() {
+				if let Some(pending_voting_period) = PendingVotingPeriod::<T>::get() {
+					VotingPeriod::<T>::set(pending_voting_period);
+				}
+			}
```
mattsse, line 465: this is not deterministic because there might be no proposal in the voting period so pending voting period won't be set until a new proposal gets closed

pallets/committee/src/lib.rs
```diff
@@ -524,7 +537,12 @@ pub mod pallet {
 				Error::<T>::InvalidVotingPeriod
 			);
 
-			VotingPeriod::<T>::set(voting_period);
+			if Proposals::<T>::iter_values().filter(|p| p.status == ProposalStatus::Active).count().is_zero() {
+				VotingPeriod::<T>::set(voting_period);
```
mattsse, line 541: I think we should always wait until the current voting period is over and use always the pending storage

mattsse (changes requested):
Instead of adjusting the voting period in `close`, we should do the migration in the `upkeep` function or in the `initialize` hook

pallets/committee/src/lib.rs
```diff
@@ -325,6 +329,13 @@ pub mod pallet {
 				proposals.retain(|hash| if let Some(votes) = Self::get_votes_for(hash) { votes.end > n } else { false })
 			});
 
+			// reset the voting period if has pending voting period
+			if let Some(pending_voting_period) = PendingVotingPeriod::<T>::get() {
+				writes = writes.saturating_add(pending_voting_period.encode().len() as Weight);
```
mattsse, line 334: why do we `encode`here? this is a single additional write, the `VotingPeriod::<T>::set`

pallets/committee/src/lib.rs
```diff
@@ -524,7 +533,7 @@ pub mod pallet {
 				Error::<T>::InvalidVotingPeriod
 			);
 
-			VotingPeriod::<T>::set(voting_period);
+			PendingVotingPeriod::<T>::set(Some(voting_period));
```
mattsse, line 536: we should add events for that as well, one for Pending with `voting_period` and when it takes effect, one for when it's set in the `upkeep` function

mattsse (changes requested):
some suggestions and needs tests to validate voting period transition

pallets/committee/src/lib.rs
clearloop: oops, considered it determined by the data length

pallets/committee/src/tests.rs
```diff
@@ -525,21 +532,24 @@ fn propose_constituent_works() {
 #[test]
 fn can_set_voting_period() {
```
clearloop, line 533: > tests to validate voting period transition

Can you describe this in more detail? this test includes the test of the range

pallets/committee/src/lib.rs
```diff
@@ -190,6 +194,12 @@ pub mod pallet {
 		/// A member has been removed
 		/// \[member_type, address]
 		RemoveMember(AccountIdFor<T>, MemberType),
+		/// A new pending voting period has been set
+		/// \[block_number]
+		NewPendingVotingPeriod(BlockNumberFor<T>),
```
mattsse, line 199: let's also add the block number when this will happen

mattsse (approved):
last nit

mattsse (approved):
lgtm

pallets/committee/src/lib.rs
```diff
@@ -522,7 +538,13 @@ pub mod pallet {
...
 			);
 
-			VotingPeriod::<T>::set(voting_period);
+			PendingVotingPeriod::<T>::set(Some(voting_period));
+			Self::deposit_event(Event::NewPendingVotingPeriod(
+				voting_period,
+				Self::get_next_voting_period_end(&frame_system::Pallet::<T>::block_number())?
+					.saturating_sub(T::ProposalSubmissionPeriod::get())
+					.saturating_sub(VotingPeriod::<T>::get()),
+			));
```
clearloop, line 547: Modified this to the `next_voting_circle` which matches the check-and-update of upkeep

mattsse (approved):
lgtm

PR #426, opened 2021-09-28
chore: document ED behaviour in add asset and saft

### LINKED ISSUE #414: Enforce existential deposit when minting PINT
## Issue summary

- `deposit_creating` may return a `PositiveImbalance` if the given amount is lower the existential deposit which is not an error, instead it should be enforced that the targeted account has the necessary balance after the deposit

-

mattsse:
## Changes

- The question raised in #414 can only happen if the caller account does not exist yet and the given nav is below ED, this is unlikely to happen, but if it does it is to the disadvantage of the caller. 
- The regular `deposit` call ensures that the pint amount is within a certain range which prevents the ED edge case entirely.

## Tests

```

```

## Issues

- Closes #414

### DIFF
--- a/pallets/asset-index/src/lib.rs
+++ b/pallets/asset-index/src/lib.rs
@@ -747,10 +747,15 @@ pub mod pallet {
 				.ok_or_else(|| ArithmeticError::Overflow.into())
 		}
 
-		/// Adds liquid asset
+		/// Adds liquid asset units
 		///
-		/// - If asset not exists, create asset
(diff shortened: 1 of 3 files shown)


clearloop (approved):
Should we also add tests for these no-op situations?

mattsse:
> Should we also add tests for these no-op situations?

good point, will add some

clearloop (approved):
docker build failed, needs to keep on watching if it caused by our docker cache or the build image of paritytech

PR #429, opened 2021-09-29
feat(committee): introduce VotingPeriodRange

### LINKED ISSUE #428: Replace `Days` type in committee with range
## Issue summary

- currently, we check the validity of a votingperdio like this `in [T::Days, T::Days * 4]`, we should replace this with a `T::ValidVotingPeriod` type that includes both thresholds and acts as a range, similar to the `StakingCap` type of the remote asset manager

-

clearloop:
## Changes

- use `VotingPeriodRange` and remove `Days` in the config of pallet committee
- updates initial VotingPeriod
-

## Tests

```
CI
```

## Issues

- Closes #428

### DIFF
--- a/pallets/committee/src/lib.rs
+++ b/pallets/committee/src/lib.rs
@@ -495,9 +495,9 @@ pub mod pallet {
 				let ty = maybe_member.take().ok_or(Error::<T>::NotMember)?;
 
 				// Check if have enough council members
-				if ty == MemberType::Constituent
-					|| Members::<T>::iter_values().filter(|m| *m == MemberType::Council).count()
-						> T::MinCouncilVotes::get()
+				if ty == MemberType::Constituent ||
+					Members::<T>::iter_values().filter(|m| *m == MemberType::Council).count() >
+						T::MinCouncilVotes::get()
 				{
 					VotingEligibility::<T>::take(&member);
 					Ok(ty)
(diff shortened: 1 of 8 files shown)


pallets/committee/src/lib.rs
```diff
@@ -516,11 +517,8 @@ pub mod pallet {
 		pub fn set_voting_period(origin: OriginFor<T>, voting_period: T::BlockNumber) -> DispatchResult {
 			T::ApprovedByCommitteeOrigin::ensure_origin(origin)?;
 
-			let days = T::Days::get();
-			let weeks = days.saturating_mul(7u32.into());
-
 			ensure!(
-				weeks < voting_period && voting_period < weeks.saturating_mul(4u32.into()),
+				T::VotingPeriodRange::min() < voting_period && voting_period < T::VotingPeriodRange::max(),
```
mattsse, line 521: this should be `[min, max]` not `]min, max[` so that min and max are still in bounds

mattsse (changes requested):
needs additional tests that update the period with the boundary values

pallets/committee/src/tests.rs
```diff
@@ -526,20 +526,19 @@ fn propose_constituent_works() {
 fn can_set_voting_period() {
```
clearloop, line 526: > needs additional tests that update the period with the boundary values

mattsse (approved):
lgtm

PR #448, opened 2021-11-10
feat: add asset tx payments

### LINKED ISSUE #167: Allow fee payment with different assets.
Since depositing xcm assets and converting them into PINT is a two step process, payment for tx in multiple currencies should be supported.
We always have at least one price available for every assets that was registered. which should be the exchange rate to charge the user. Default currency is PINT token.

Ideally this should be done with a custom pallet that implements `TransactionPayment`

Ref #143 #88 
_Originally posted by @mattsse in https://github.com/ChainSafe/PINT/issues/146#issuecomment-874642658_

mattsse:
## Changes

- Allows to charge for transactions with any asset

## Tests

```

```

## Issues

- Closes #167


clearloop (approved):
lgtm

PR #473, opened 2021-12-14
chore(substrate): upgrade v0.9.13

clearloop:
## Changes

- upgrade PINT to v0.9.13

## Tests

```
CI
```

## Issues

- Closes #471

PR #476, opened 2021-12-21
feat(fee): update fee charge in runtime

### LINKED ISSUE #472: Improve XCM weight Handling
## Issue summary

- Currently we use `UnitPerSecond`, to improve this we should separate this to something like `pint_per_sec` and test accordingly
- The `Unit` weight is also configured too high currently

-

clearloop:
## Changes

based on #473 

- update xcm transaction fees

## Tests

```
CI
```

## Issues

- Closes #472


runtime/common/src/payment.rs
```diff
@@ -24,3 +27,21 @@ where
 		NAV::asset_equivalent(balance, asset_id)
 	}
 }
+
+fn base_tx_in_pint() -> Balance {
+	1 / 1000
+}
```
clearloop, line 33: using `1 / 1000` as basic fee

mattsse (approved):
🎉

